PR #51, opened 2017-06-20
create profile component

ahdinosaur:
- [x] app.TextField component and story
- [x] app.TextAreaField component and story
- [x] app.ImageField component and story
  - [x] use https://github.com/mosch/react-avatar-editor
  - [x] add file input (to click and pick file)
  - [x] test drag and drop file works
  - [x] add zoom slider
  - [x] store image data in form state
- [x] agents.Profile component and story
  - [x] agents.Profile editing and display modes
  - [x] agents.ProfileIcon component and story


agents/components/Avatar.js
```diff
@@ -16,7 +16,7 @@ function Avatar (props) {
   const {
     size,
     styles,
-    agent: { profile: { name, image } }
+    image
```
ahdinosaur: hmm... i'm not sure how i feel about this change.

my thinking was that the components we are building are not strictly generic, they are re-usable but specific to the data model concepts in Cobuy. so since an "agent" is a first class data model concept, maybe we should build our components to accept them as arguments directly, to avoid all sorts of destructuring and restructuring code. or, what if we want to later include the agent name when the avatar is large. or put a star icon in the avatar if they are the admin of the current group.

but maybe i'm being silly.

gregorykan: yeah, i had some questions around levels here - let's discuss after lunch!

PR #56, opened 2017-06-22
[wip] invite members story

### LINKED ISSUE #55: story: invite members component
-

gregorykan:
closes #55 

- [x] member invites component and story
  - [x] automate addition of new field when first/previous field is being populated


agents/components/MemberInvites.js
```diff
@@ -0,0 +1,59 @@
...
+import { flow } from 'lodash'
+import { TextField } from 'redux-form-material-ui'
+
+import styles from '../styles/MemberInvites'
+import Button from '../../app/components/Button'
+
+function renderMembers ({ fields, meta: { error, submitFailed } }) {
+  return (
+    <div>
+      <Button type='button' onClick={() => fields.push({})}>Add Member</Button>
```
ahdinosaur: so if we want to do the "next member field always showing" approach, then rather than add a button to create an empty member, we'll want to add an extra set of fields at the bottom of this render (as if it were an empty member), where pass in `onChange` to each of these "next member" `<Field />` so when you start typing it will `push` the next member, which will cause the next render to include that as a regular field and the next empty set of fields will be ready.

agents/components/MemberInvites.js
```diff
@@ -1,17 +1,27 @@
...
-import { Field, FieldArray, reduxForm as connectForm } from 'redux-form'
+import { Field, FieldArray, formValueSelector, reduxForm as connectForm } from 'redux-form'
 import { flow } from 'lodash'
 import { TextField } from 'redux-form-material-ui'
 
 import styles from '../styles/MemberInvites'
 import Button from '../../app/components/Button'
 
-function renderMembers ({ fields, meta: { error, submitFailed } }) {
+function renderMembers ({ fields, meta: { error, submitFailed }, formProps }) {
```
gregorykan: @ahdinosaur so this works as we expect, but there are a few things i'm not entirely sure/comfortable about:

- scrappy nested conditional logic to check that the last row has values in it
- react does not like the fact that this is taking place within the render cycle
  - i thought about retreating to using componentWillReceiveProps, but then was unsure how i would gain access to `fields` - which seems to belong to the `FieldArray` component - in order to push a new one

whaddya think?

PR #57, opened 2017-06-23
alternative approach to member invites new row

ahdinosaur:
https://github.com/root-systems/cobuy/pull/56#discussion_r123433618
https://github.com/root-systems/cobuy/pull/56/files/9c8dbfd2afff43a455a06152fd5b2698d179ff60..85e5d7815ca518d0060f907b9b8689e87ce55061#r123647839

### DIFF
--- a/agents/components/MemberInvites.js
+++ b/agents/components/MemberInvites.js
@@ -5,76 +5,95 @@ import { connect as connectFela } from 'react-fela'
 import { Field, FieldArray, formValueSelector, reduxForm as connectForm } from 'redux-form'
 import { flow } from 'lodash'
 import { TextField } from 'redux-form-material-ui'
+import MuiTextField from 'material-ui/TextField'
 
 import styles from '../styles/MemberInvites'
 import Button from '../../app/components/Button'
 
-function renderMembers ({ fields, meta: { error, submitFailed }, formProps }) {
-  const { memberVals } = formProps
-  // TODO: currently this is an anti-pattern as it occurs within the render cycle
-  // TODO: mikey's idea was to not push state until the first edit to the empty row
-  if (memberVals) {
-    const memberKeys = Object.keys(memberVals)
-    if (memberKeys.length > 0) {
-      const lastMember = memberVals[memberKeys[memberKeys.length - 1]]
-      if (Object.keys(lastMember).length > 0) {
-        fields.push({})
-      }
+class MemberInviteFields extends React.Component {
+  render () {
+    const { fields } = this.props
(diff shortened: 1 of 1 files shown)


agents/components/MemberInvites.js
```diff
@@ -5,76 +5,95 @@ import { connect as connectFela } from 'react-fela'
 import { Field, FieldArray, formValueSelector, reduxForm as connectForm } from 'redux-form'
 import { flow } from 'lodash'
 import { TextField } from 'redux-form-material-ui'
+import MuiTextField from 'material-ui/TextField'
```
gregorykan, line 8: how come we use the standard mui text field in addition to the redux-form one ?

agents/components/MemberInvites.js
```diff
@@ -5,76 +5,95 @@ import { connect as connectFela } from 'react-fela'
...
+    const { fields } = this.props
+    return (ev, value) => {
+      fields.push({ [fieldName]: value })
+      const { members } = this.refs
+      // HACK (mw) how can we clean this up?
+      // keep focus after we push next member field
+      // because react elements will change underneath.
+      // also there's enough of a stutter that when
+      // typing you will lose some characters.
+      setTimeout(() => {
```
gregorykan, line 69: i wonder why we lose focus with this approach, and not in mine?

agents/components/MemberInvites.js
ahdinosaur: the "next row" is not yet part of the redux form state (as part of this alternative approach), so i felt it was best to represent it outside of the redux form components. i did try this as a redux form `<Field />`, but it felt weird to guess what the required `name` prop would be (`members[${fields.length}]`).

agents/components/MemberInvites.js
ahdinosaur: in your approach, you pre-emptively add an empty field to the redux form state, then let it render as normal (even though it's empty). this means the react element tree works as normal, since the "empty row" is part of the `map(fields, renderField)` like any other row.

in my approach, i separate the "empty row" from the normal rows (for example, i don't have a remove button here because how can you remove something that doesn't exist yet :stuck_out_tongue_winking_eye:), so the react element tree is different, which causes react to change the underlying dom elements.

if we can't figure out this focus problem in a graceful way, i think we should continue with your approach but find a way to gracefully handle the "next empty row" being part of the state. (we need to handle this whenever we "save" the state, at least so the empty row state is isolated to this component.)

ahdinosaur:
closing

PR #60, opened 2017-06-26
Updates the db migration symlinks

danalexilewis:
previously this pointed to dogstack-agents as a seperate repo, this points it to node-modules/dogstack-agents an npm dependancy.


ahdinosaur:
churrrrrr! :cat:

PR #81, opened 2017-07-05
Chore/implement intl messages

### LINKED ISSUE #71: implement i18n on existing components

gregorykan:
closes https://github.com/root-systems/cobuy/issues/71


app/helpers/classifyIntlMessage.js
```diff
@@ -0,0 +1,11 @@
+import React from 'react'
+
+export default function (className) {
```
ahdinosaur: can we link to the issue that describes why this is necessary, as in the `notfix` comment?

ahdinosaur: also, given how much this was used in this pull request, how do we feel about providing a wrapped `FormattedMessage` as well, using this?

```js
import { FormattedMessage: OgFormattedMessage } from 'react-intl'
import { isNil } from 'ramda'

function FormattedMessage (props) {
  return isNil(props.className)
    ? <OgFormattedMessage {...props} />
    : <OgFormattedMessage {...props} {...classifyIntlMessage(props.className) } />
}
```

app/locales/en.json
```diff
@@ -1,5 +1,30 @@
 {
   "app.name": "Cobuy",
   "agents.nameLabel": "name",
-  "agents.descriptionLabel": "description"
+  "agents.descriptionLabel": "description",
```
ahdinosaur, line 4: woo!

app/locales/en.json
```diff
@@ -1,5 +1,30 @@
...
+  "agents.signIn": "sign in",
+  "agents.welcome": "hey, welcome to cobuy!",
+  "agents.signInWith": "sign in with...",
+  "agents.saveAvatar": "save avatar",
+  "agents.editAvatar": "edit avatar",
+  "supply.removeOffering": "remove offering",
+  "supply.addOffering": "add offering",
+  "supply.quantity": "quantity",
+  "supply.unit": "unit",
+  "supply.litres": "litres",
```
ahdinosaur, line 25: is `en` American English or British English? maybe also an opportunity to add `en-US` and `en-GB` locales, for whichever is not our default (which we should document somewhere, otherwise i'll write American English as my default).

ahdinosaur: oh, i realize this also affects our message ids.

app/helpers/classifyIntlMessage.js
ahdinosaur: also also, now i'm thinking this should live in `cobuy/lib/`, where re-usable code that has nothing to do with our app goes, which should really become a new module or a pull request to an existing module.

ahdinosaur (approved):
legit as

gregorykan:
- [x] abstract formatted message being passed child function into lib folder
  - [x] link to issue regarding formatted message not taking className as props directly, instead taking child function
  - [x] refactor code to use formatted message wrapper
- [x] specify 'developer language' as british english in the readme
- [x] specify default locale as 'en-gb', also provide 'en-us'

agents/components/SignIn.js
```diff
@@ -8,6 +8,7 @@ import FlatButton from 'material-ui/FlatButton'
 import RaisedButton from 'material-ui/RaisedButton'
 import FontIcon from 'material-ui/FontIcon'
 
+import { FormattedMessage } from '../../../cobuy/lib/Intl'
```
ahdinosaur: too many `../`, should be `../../lib/Intl`

app/locales/en-us.json
```diff
@@ -0,0 +1,3 @@
+{
+  "supply.litres": "liters"
```
ahdinosaur, line 2: yay! :us:

PR #82, opened 2017-07-06
Feature/member intent field

### LINKED ISSUE #49: component: individual intent casting
creating the individual intent casting view so it can be tested and tweaked

https://github.com/root-systems/cobuy/issues/20

gregorykan:
closes https://github.com/root-systems/cobuy/issues/49


ahdinosaur:
base functionality works:

cobuy-intent

ahdinosaur:
much better meow:

cobuy-intent2

gregorykan:
@ahdinosaur dragging the minimum to zero results in `Uncaught Error: new BigNumber() not a number: undefined` - can't imagine this being difficult to resolve tho?

overall this feels great! i wonder a bit about the way we label the limits/bounds and axes to make it more coherent for users. but perhaps we shouldn't worry about that yet.

PR #84, opened 2017-07-11
setup veel

### LINKED ISSUE #85: style profile component

gregorykan:
- [x] setup veel
- [x] style profile component

closes https://github.com/root-systems/cobuy/issues/85

### DIFF
--- a/app/themes/base.js
+++ b/app/themes/base.js
@@ -69,5 +69,5 @@ export default {
   rem: '1rem',
   fontFamily: 'Roboto, sans-serif',
   borderRadius: '2px',
-  zIndexMax: 100
+  zIndexMax: 1000
 }
--- a/app/themes/base.js
+++ b/app/themes/base.js
@@ -33,25 +33,29 @@ export default {
     12, 16, 18, 24, 36, 72
   ],
   colors: {
-    teal500,
-    teal600,
-    teal700,
-    deepPurple500,
-    deepPurple600,
-    deepPurple700,
-    grey50,
-    grey100,
-    grey200,
-    grey300,
-    grey400,
-    grey500,
-    grey600,
-    grey700,
-    grey800,
-    grey900,
-    darkBlack,
-    fullBlack,
-    white
+    primary1: teal500,
+    primary2: teal600,
+    primary3: teal700,
+    accent1: deepPurple500,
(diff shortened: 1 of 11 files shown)


app/themes/base.js
```diff
@@ -17,41 +15,30 @@ import {
...
+  breakpoints: [
+    32, 48, 64
+  ],
+  space: [
+    0, 6, 12, 18, 24
+  ],
+  fontSizes: [
+    12, 16, 18, 24, 36, 72
+  ],
+  colors: {
```
ahdinosaur, line 35: i think these colors should still be "semantic" names, like `primary1` rather than `teal500`. i'd expect this change was only to nest the semantic names within the `colors` object, so `colorPrimary1` becomes `colors.primary1`, same for `accent1`, `text`, `alternativeText`, `canvas`, `border`, `shadow`.

ahdinosaur: although the grey / white / black colors are probably all good as you did.

app/themes/base.js
```diff
@@ -61,21 +48,18 @@ export default {
...
-  colorShadow: fullBlack,
-  spacingNone: 0,
-  spacingExtraSmall: '.25rem',
-  spacingSmall: '.5rem',
-  spacingMedium: '1rem',
-  spacingLarge: '2rem',
-  spacingExtraLarge: '4rem',
-  spacingExtraExtraLarge: '8rem',
-  spacingExtraExtraExtraLarge: '16rem',
-  borderRadius: '2px',
```
ahdinosaur: do we want to get rid of these parts of the theme?

- `borderRadius`
- `zIndexMax`
- `em`
- `rem`

ahdinosaur (changes requested):

app/themes/base.js
gregorykan: good point re: retaining semantic names - i thought about this and wasn't sure! i agree with you as a mere list of colours (without context) wouldn't be very useful in a theme. that context-less list is how the styled-system person did it so i hesitated

app/themes/base.js
gregorykan: nope! will add those back in

ahdinosaur (approved):

PR #89, opened 2017-07-12
Chore/style member invites component

### LINKED ISSUE #70: style member invites component

gregorykan:
closes https://github.com/root-systems/cobuy/issues/70

### DIFF
--- a/agents/components/MemberInvites.js
+++ b/agents/components/MemberInvites.js
@@ -81,6 +81,12 @@ function MemberInvites (props) {
   const { styles } = props
   return (
     <form className={styles.container}>
+      <p className={styles.intro}>
+        <FormattedMessage
+          id='agents.memberInvites'
+          className={styles.labelText}
+        />
+      </p>
       <Field
         name='groupName'
         type='text'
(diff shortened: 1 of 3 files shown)


agents/components/MemberInvites.js
```diff
@@ -81,17 +85,25 @@ function MemberInvites (props) {
...
-          }
-      />
+      <p className={styles.intro}>
+        <FormattedMessage
+          id='agents.memberInvites'
+          className={styles.labelText}
+        />
+      </p>
+      <div className={styles.groupNameContainer}>
+        <Field
```
ahdinosaur: hey, i don't think we should have group name as part of the member invites component, this is handled by the group agent profile component, right?

ahdinosaur (commented):
nice!

two main comments:

- it's not responsive (see screenshot in next comment)
- i don't think we should ask the group name here as i expect we use the profile component to edit the group name, etc

ahdinosaur:
not responsive:

cobuy-member-invites

should it be?

ahdinosaur:
screenshot:

cobuy-member-invites1

@gregorykan would be appreciated if you could post screenshots or gifs as part of the pull request, since it helps document our progress for stakeholders and future archaeologists (like ourselves). for gifs i use https://github.com/phw/peek. :smile:

agents/components/MemberInvites.js
gregorykan: @ahdinosaur good point ! will remove

gregorykan:
responsive-member-invites

PR #91, opened 2017-07-12
revert base theme to use pre-existing rem values

ahdinosaur:
8009781f391e300de548dabf6c27e619570a9789

@gregorykan, whatcha think bout this?


gregorykan:
@ahdinosaur makes sense to me ! think i took the adaptation of styled-system's theme a bit too literally

PR #100, opened 2017-07-13
Feature/navigation component

### LINKED ISSUE #95: component: Navigation
should be something like: https://negomi.github.io/react-burger-menu/, but with material ui.

- NavigationToggle
  - [font awesome hamburger icon](http://fontawesome.io/icon/bars/) (we already have fa installed)
- NagivationBar
  - [material ui app bar](http://www.material-ui.com/#/components/app-bar) (NavigationBar)
- NavigationDrawer
  - [material ui drawer](http://www.material-ui.com/#/components/drawer)
- NavigationMenu
  - shows "LogOut", "Dashboard" buttons (for now)
  - split menu into sections with [material ui subheader](http://www.material-ui.com/#/components/subheader) or [material ui divider](http://www.material-ui.com/#/components/divider)

gregorykan:
closes https://github.com/root-systems/cobuy/issues/95

### DIFF
--- a/app/components/Navigation.js
+++ b/app/components/Navigation.js
@@ -1,80 +1,73 @@
 import React from 'react'
 import { connect as connectFela } from 'react-fela'
-import { pipe } from 'ramda'
+import { not } from 'ramda'
 import AppBar from 'material-ui/AppBar'
 import Drawer from 'material-ui/Drawer'
 import MenuItem from 'material-ui/MenuItem'
 import Divider from 'material-ui/Divider'
+import { withState, withHandlers, compose } from 'recompose'
 
 import styles from '../styles/Navigation'
 import { FormattedMessage } from '../../lib/Intl'
 
-class Navigation extends React.Component {
-  constructor (props) {
-    super(props)
-    this.state = {
-      drawerOpen: false
-    }
-  }
+function Navigation (props) {
+  const { styles, isDrawerOpen, toggleDrawer } = props
 
-  handleDrawerToggle = () => {
-    this.setState({ drawerOpen: !this.state.drawerOpen })
-  }
-
-  render () {
-    const { styles } = this.props
-    return (
-      <div>
-        <AppBar
-          title={
-            <FormattedMessage
-              id='app.name'
-              className={styles.labelText}
-            />
+  return (
+    <div>
+      <AppBar
+        title={
+          <FormattedMessage
+            id='app.name'
+            className={styles.labelText}
+          />
+        }
+        onLeftIconButtonTouchTap={toggleDrawer}
+      />
+      <Drawer open={isDrawerOpen}>
+        <MenuItem
+          leftIcon={
+            <i className="fa fa-bars" aria-hidden="true"/>
           }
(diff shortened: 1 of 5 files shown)


app/components/Navigation.js
```diff
@@ -0,0 +1,80 @@
...
+import Divider from 'material-ui/Divider'
+
+import styles from '../styles/Navigation'
+import { FormattedMessage } from '../../lib/Intl'
+
+class Navigation extends React.Component {
+  constructor (props) {
+    super(props)
+    this.state = {
+      drawerOpen: false
```
ahdinosaur: i'm a fan of `is*` boolean props, so `isDrawerOpen`. helps disambiguate from thinking that this is a function which opens the drawer. what do you think?

app/components/Navigation.js
```diff
@@ -0,0 +1,80 @@
...
+import MenuItem from 'material-ui/MenuItem'
+import Divider from 'material-ui/Divider'
+
+import styles from '../styles/Navigation'
+import { FormattedMessage } from '../../lib/Intl'
+
+class Navigation extends React.Component {
+  constructor (props) {
+    super(props)
+    this.state = {
```
ahdinosaur: i know i mentioned this before, but what do you think about using `recompose` to handle stateful or lifecycle components? while i did say "functional purity" can be a trap, i am enjoying using `recompose` so far.

so this would be:

```js
import { connect as connectFela } from 'react-fela'
import { withState, withHandlers } from 'recompose'
import { pipe, not } from 'ramda'

import styles from '../styles/Navigation'

const Navigation = (props) => {
  // render...
}

export default pipe(
  connectFela(styles),
  withState('isDrawerOpen', 'setDrawerOpen', false),
  withHandlers({
    toggleDrawer: ({ setDrawerOpen }) => () => setDrawerOpen(not)
  })
)(Navigation)
```

ahdinosaur (commented):
yay, this is cool.

app/components/Navigation.js
gregorykan: yeaaaahhh nice ! i wanted to ask you how to do this actually - will give it a go!

app/components/Navigation.js
gregorykan: A++ i agree

app/components/Navigation.js
```diff
@@ -1,80 +1,73 @@
...
+        </MenuItem>
+        <Divider />
+      </Drawer>
+    </div>
+  )
 }
 
-export default pipe(
-  connectFela(styles)
+export default compose(
```
ahdinosaur: good call on using `recompose.compose` :+1:

app/components/Navigation.js
```diff
@@ -1,80 +1,73 @@
...
 import { FormattedMessage } from '../../lib/Intl'
 
-class Navigation extends React.Component {
-  constructor (props) {
-    super(props)
-    this.state = {
-      drawerOpen: false
-    }
-  }
+function Navigation (props) {
```
ahdinosaur: beautiful! :smiley_cat:

PR #104, opened 2017-07-18
initial getters for tasks, and tests

gregorykan:


danalexilewis:
Over the moon that we are doing this - very much in line with how I was thinking of organising getters for interfaces in https://github.com/agentlewis/breeder. Currently, it doesn't but this will work really nicely for it.

PR #110, opened 2017-07-20
add implicit child plans to getTaskPlans

ahdinosaur:
an "implicit child plan" is where the parent task has a child recipe but no corresponding child plan.

---

@gregorykan: i think we need to re-visit task getters in general. when i wrote the TaskWorker component, i made a first pass at the data structure returned: https://github.com/root-systems/cobuy/blob/c44d2726ec559fa5eb7c8bbcfbee6b21737b2c71/tasks/stories/TaskWorker.js#L17-L37, but now since TaskWorks no longer need a corresponding TaskPlan, i think we need to re-think this.

then there's this case, where i took the "implicit child plan" approach, so we can create a single parent plan to count for all the child plans, but i'm not sure this is the best approach.

also i wonder if the role of the getter is to turn all these separate Recipe, Plan, and Work objects into a single conceptual "Task".
 
hmm.

### DIFF
--- a/tasks/getters/getTaskPlans.js
+++ b/tasks/getters/getTaskPlans.js
@@ -50,6 +50,28 @@ const getTaskPlanTree = createSelector(
         .filter(complement(isNil))
     }
 
+    // default any missing child plans where child recipes exist
+    Object.keys(taskPlansById).forEach(taskPlanId => {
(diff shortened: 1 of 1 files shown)


ahdinosaur:
actually wait, what is the route to complete an implicit task plan? we might need to create all child task plans. :sweat_smile:

PR #116, opened 2017-07-25
backend to create all the respective objects and relations on order start

gregorykan:


agents/components/Profile.js
```diff
@@ -1,118 +1,109 @@
...
-            }
-          </RaisedButton>
-        </div>
-      </form>
-    )
-  }
-
-}
-
-Profile.propTypes = {
```
danalexilewis, line 101: @gregorykan whats the thinking around proptypes? are we not going to use them at all?

and @ahdinosaur I am still curious and keen on an implementation of types on the apis of modules. Is the thinking that components shouldn't have a type api?

ahdinosaur: i think we've decided to not use them. but also still keen on a type api, just haven't made it a priority in our sprints to add type schemas, which maybe we should do.

ahdinosaur (approved):

agents/components/Profile.js
ahdinosaur: here's an issue for this meow: https://github.com/root-systems/cobuy/issues/118

PR #122, opened 2017-07-31
Feature/landing

### LINKED ISSUE #28: page: landing
- [x] dumb component (maybe add to storybook)
- [x] make it work with the app
- [ ] screenshot/gif grab

sarah-arrrgh:
closes #28


app/styles/home.js
```diff
@@ -1,9 +1,106 @@
...
+        width: '100%',
+        height: '100%',
+        opacity: .6,
+        zIndex: -1,
+        backgroundColor: theme.colors.canvas,
+        backgroundImage: 'url("/images/cobuy-bg-sml-1080.jpg")',
+        backgroundSize: 'cover'
+      },
+      [`@media (min-width: ${theme.breakpoints.desktop})`]: {
+        content: '""',
```
ahdinosaur: `content` is specific to a `::before` element. is this....

oh, i got confused by the indentation, this is all within a `::before`. ⚡️

ahdinosaur: the properties that are shared between media queries could be taken out from the media query and put in the parent level.

app/themes/base.js
```diff
@@ -24,32 +24,47 @@ import {
...
 export default {
-  breakpoints: [
-    '13.6em',
-    '30.4em',
-    '47.2em',
-    '64em',
-    '97.6em'
-  ],
+  breakpoints: {
+    mobile: '30em', // 480px,
```
ahdinosaur, line 28: nice!

app/themes/base.js
```diff
@@ -24,32 +24,47 @@ import {
...
-    '8rem',
-    '16rem',
+    '0rem', // [0] SR: to override default allofthespace with noneofthespace
+    '0.25rem', // [1]
+    '0.5rem', // [2]
+    '1rem', // [3]
+    '2rem', // [4]
+    '4rem', // [5]
+    '8rem', // [6]
+    '12rem', // [7]
```
ahdinosaur: i see we're moving away from the exponential ratio-based spacing from [tachyon css](http://tachyons.io/docs/layout/spacing/). what's the motivation for the specific spacing values chosen? what was problematic about using the existing spacing scale?

app/themes/base.js
```diff
@@ -24,32 +24,47 @@ import {
...
   fontSizes: [
-    '0.875rem',
-    '1rem',
-    '1.25rem',
-    '1.5rem',
-    '2.25rem',
-    '3rem',
-    '5rem',
-    '6rem'
+    '0.8rem', // [0] 10pt
```
ahdinosaur: same comment as with the spacings. these font size values were chosen from [tachyon css](http://tachyons.io/docs/typography/scale/), why should we change to using these new values?

ahdinosaur (commented):
sweet! 🎉 

my only serious concern is about changing our font size and spacing theme values to diverge from [tachyons](http://tachyons.io/docs/). i'm open to being convinced of a new design paradigm, but i don't think we should arbitrarily diverge without adopting a new, equally clear and consistent paradigm.

also, since this pull request includes our first large image, i'd be keen to use [Git Large File Storage](https://git-lfs.github.com). i had documentation about this in the previous Cobuy version (catstack era), but i guess i removed that during the composting. i'm happy to do this during the merge if it's confusing.

app/themes/base.js
sarah-arrrgh: It's based on typographic font scales, there's a comment on it in the landing issue @ahdinosaur , about the 7th comment down

app/styles/home.js
sarah-arrrgh: will update indentation!

app/themes/base.js
ahdinosaur: cool. can you add a comment in the code linked to the source for the font scale?

sarah-arrrgh: i just did a google image search for those images, but will see if i can find a good article to back it up

sarah-arrrgh: http://spencermortensen.com/articles/typographic-scale/
maybe?

sarah-arrrgh: and
https://blog.madewithenvy.com/responsive-typographic-scales-in-css-b9f60431d1c4

sarah-arrrgh:
I tried to do the Git LFS thing but I can't figure out how to get past the first step :( can you please do it @ahdinosaur ?

app/themes/base.js
ahdinosaur: this is awesome, thanks for finding and sharing!

with this in mind, i'm keen to use the "pentatonic scale" (doubles every fifth step) as described, plus using `em` instead of `rem` when using this scale as recommended in the second article.

```
2^(5/5)      ->  2
2^(4/5)      ->  1.7411
2^(3/5)      ->  1.51572
2^(2/5)      ->  1.31951
2^(1/5)      ->  1.14869
2^(-1/5)     ->  0.87055
2^(-2/5)     ->  0.75785
2^(-3/5)     ->  0.65975
2^(-4/5)     ->  0.57435
2^(-5/5)     ->  0.5
```

so if we follow the second article's advise about using a single up/down scale and nesting `em`s, we might use:

```js
const fontSizes = {
  '-5': 0.50000,
  '-4': 0.57435, 
  '-3': 0.65975,
  '-2': 0.75785,
  '-1': 0.87055,
  0: 1.00000,
  1: 1.14869,
  2: 1.31951,
  3: 1.51572,
  4: 1.74110,
  5: 2.00000,
  // continue with formula:
  //   multiplier ^ (step / interval)
  //   2 ^ (step / 4)
]
```

or more like we did it before:

```js
const fontSizes = {
  0.50000, // [0]
  0.57435, // [1]
  0.65975, // [2]
  0.75785, // [3]
  0.87055, // [4]
  1.00000, // [5]
  1.14869, // [6]
  1.31951, // [7]
  1.51572, // [8]
  1.74110, // [9]
  2.00000, // [10]
  // continue with formula:
  //   multiplier ^ (step / interval)
  //   2 ^ (step / 4)
]
```


happy for this to be a different pull request, will start a conversation about this.

PR #131, opened 2017-08-15
fix `redux-form-validators` error with a fork

ahdinosaur:
https://github.com/ahdinosaur/redux-form-validators

fixes es module dependency error when run as node

```
/home/thatsmith/github/cobuy/node_modules/redux-form-validators/src/index.js:1
import React from 'react'
^
ParseError: 'import' and 'export' may appear only with 'sourceType: module'
```

@NotThatSmith


michael-smith-nz:
:tada: It works :tada:

PR #139, opened 2017-08-18
Task completion

### LINKED ISSUE #126: clicking "done" creates a task work to mark task completion
moving out from #120 

job story: when i am done with a task and want to mark it as complete, i click "done" and a TaskWork is created to reference the associated TaskRecipe (and optional TaskPlan).

screenshot_20170816_144226

this task should then show as "completed" on any page referencing it.

michael-smith-nz:
fixes #126

### DIFF
--- a/tasks/containers/TaskWorker.js
+++ b/tasks/containers/TaskWorker.js
@@ -1,7 +1,7 @@
 import h from 'react-hyperscript'
 import { isNil, pipe, filter, keys, length, gte, propEq, not } from 'ramda'
 import { connect as connectFeathers } from 'feathers-action-react'
-import { compose } from 'recompose'
+import { compose, lifecycle, withState, withHandlers } from 'recompose'
 import { push } from 'react-router-redux'
 
 import { actions as taskPlans } from '../dux/plans'
@@ -79,9 +79,26 @@ export default compose(
 
       return false
     }
+  }),
+  withState('createTaskWorkerCid', 'editTaskWorkerCid', null),
+  withHandlers({
+    setTaskWorkerId: ({ editTaskWorkerCid }) => (cId) => editTaskWorkerCid(cId)
+  }),
+  lifecycle({
+    componentWillReceiveProps: function (nextProps) {
+      if (!isNil(nextProps.createTaskWorkerCid) &&
+        nextProps.feathersData[nextProps.createTaskWorkerCid].isReady &&
+        isNil(nextProps.feathersData[nextProps.createTaskWorkerCid].error)
+      ) {
+        const { parentTaskPlan } = nextProps.taskPlan
+        const nextRoute = isNil(parentTaskPlan)
+          ? '/' : `/tasks/${parentTaskPlan.id}`
+        nextProps.actions.router.push(nextRoute)
+      }
+    }
   })
 )(props => {
-  const { taskPlan, currentAgent, actions } = props
+  const { taskPlan, currentAgent, actions, setTaskWorkerId } = props
 
   return h(TaskWorker, {
     taskPlan,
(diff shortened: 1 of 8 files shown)


tasks/containers/TaskWorker.js
```diff
@@ -99,6 +100,24 @@ export default compose(
...
+    const taskWork = {
+      taskPlanId: taskPlan.id,
+      taskRecipeId: taskplan.taskRecipeId,
+      workerAgentId: currentAgent.id,
+      params: {
+        // contextAgentId: group.id
+      }
+    }
+
+    actions.taskWorks.create(taskWork)
```
michael-smith-nz: Is this creation not async? I tried a callback and promise with no luck.

ahdinosaur: we're using [redux-observable](https://redux-observable.js.org/) in [feathers-action](https://github.com/ahdinosaur/feathers-action), so it's async but currently doesn't return a promise it [returns a cancellation token](https://github.com/ahdinosaur/feathers-action-react/blob/master/index.js#L51-L56).

the pattern currently is you subscribe to everything you need upfront, so that when things change you're automatically subscribed. in this case, we want to be subscribed to all task works referenced by the current task plan, so that when we create a new task work we'll be immediately notified. this subscription stuff is the `query` and `shouldQueryAgain` functions in any container using `connectFeathers`.

i figured out how we could return a promise though: if we implement a [ready action](https://github.com/ahdinosaur/feathers-action/commit/24180dc526952c87121a012e1a7b66495ccfd16d) for all methods, then we could fire off the action and resolve or reject a promise based on ready or error. but i haven't done this, i've only recently had the idea.

tasks/containers/TaskWorker.js
```diff
@@ -79,14 +79,32 @@ export default compose(
```
michael-smith-nz, line 79: I'm getting an infinite `taskWork` request here because this is always returning true and I'm  struggling to find the reason why.

@ahdinosaur @iainkirkpatrick If either of you two gets a chance to look at this that'd be rad as I'm feeling blocked. Otherwise, coming back to it with fresh eyes tomorrow might be enough.

This branch also depends on this file to be changed https://github.com/ahdinosaur/feathers-action/pull/20/files to work

ahdinosaur: hmmmm...

iainkirkpatrick: which part always returns true @NotThatSmith ? if you're online in the next 40 mins or so i can take a look at it with you

michael-smith-nz: @iainkirkpatrick Woops, that preview didn't work. This line https://github.com/root-systems/cobuy/blob/feature/task-completion/tasks/containers/TaskWorker.js#L78

Turns out https://github.com/ahdinosaur/feathers-action/pull/20 Didn't work how I intended anyway so might have to rethink things.

ahdinosaur: @NotThatSmith in case you aren't already aware (i wasn't for a while): as a general debugging tip for this situation, i recommend using `debugger` (add a line with just `debugger`). it's like `console.log('value', value)`, but better since it halts execution and you can see all the local variables at the time. 🔍  then you can halt and play it for every time it reaches the questionable moment and see what is happening there.

ahdinosaur: oh shit, i might have figured this out.

this line is bad: https://github.com/root-systems/cobuy/blob/feature/task-completion/tasks/containers/TaskWorker.js#L139

it should be `gte(__, 1)`, where `__` is a [special placeholder](http://ramdajs.com/0.21.0/docs/#__) when you need to curry in a particular order. the reason is because [`gte`](http://ramdajs.com/0.21.0/docs/#gte) expects arguments in `x > y` order, which means we are currying `x` as `1`, so it only works when it shouldn't work. (oops!)

another tip that helped me here (i found this while working on #141, made the same mistake a second time) was using [`tap`](http://ramdajs.com/0.21.0/docs/#tap) inside a ramda `pipe` like this.

michael-smith-nz: Yah! That works, thanks @ahdinosaur! 😀

iainkirkpatrick (approved):
looks rad to me - tasks complete when i finish em - might be worth @ahdinosaur checking the PR on `feathers-action`, though we can always fix that later

iainkirkpatrick:
merging now `feathers-action` PR 21 has been merged

Issue #32, opened 2017-06-19
page: create my profile


ahdinosaur:
- [ ] paper sketch (upload photo)
- [x] dumb component (maybe add to storybook)
- [x] add it to `dogstack-agents`
- [ ] make it work in the app

needs to be a personal task assigned when someone registers as a new user.

closes #140

sarah-arrrgh:
End of day report:

- made a (temporary) button on the dashboard to create a create profile task plan. wired it all up magically.

TODO: 

- [ ] fix git lfs thing with @ahdinosaur 
- [ ] wire up view/edit profile component
- [ ] probably other things... @iainkirkpatrick ?

sarah-arrrgh:
TODO Thursday arvo:

- [x] set up props for CreateProfileTask Container/Component to go through to Profile component
- [ ] add more fields to the person profile 
- [ ] convert button to hook

sarah-arrrgh:
End of day report:

- generic Profile component rendering as the component of a CreateProfileTask
- looked at adding more fields to the Profile for a Person, but decided in consultation with @ahdinosaur that it'd be good to have deliberate design on which fields make sense before developing
- investigated how to add a hook to `dogstack-agents` for creating the CreateProfileTask after a Person is created - needs more work

TODO:
- [ ] chat to Mikey more about hooks on `dogstack-agents`
- [ ] if iain gets a chance to try more hooks work, then explain to sarah on Monday what progress has been made

iainkirkpatrick:
got a hook working tonight - will solicit feedback tomorrow / monday :)

PR #140, opened 2017-08-20
Feature/create my profile

iainkirkpatrick:
adds a Task for the current Agent to create their personal profile
Task is created as a hook on the Agents service - a hooks set that is applied onto the service from `dogstack-agents`

no additional fields for the personal Profile compared to the current group Profile, after discussion it was decided to deliberately design these later


agents/services/hooks/agents.js
```diff
@@ -0,0 +1,21 @@
+module.exports = function () {
+  const app = this
+  app.service('agents').hooks(hooks)
```
ahdinosaur: 👍

server.js
```diff
@@ -2,7 +2,8 @@ const services = [
   require('dogstack-agents/service'),
   require('./tasks/services/plans'),
   require('./tasks/services/works'),
-  require('./ordering/services/orders')
+  require('./ordering/services/orders'),
+  require('./agents/services/hooks/agents')
```
ahdinosaur: i reckon this should just be `./agents/services/agents.js`

tasks/components/CreateProfileTask.js
```diff
@@ -0,0 +1,18 @@
...
+import Profile from '../../agents/components/Profile'
+
+export default (props) => {
+  const { actions, currentAgent } = props
+
+  if(isNil(currentAgent)) {
+    return null
+  }
+
+  return <Profile
```
ahdinosaur: hyperscript?

ahdinosaur: obviously need buy-in from the rest of the team, but my current stack opinion is to use `hyperscript` and `recompose` to make pure functional components using only functions, on any new code or when we update old code.

why hyperscript? it makes us less dependent on a build step (since JSX will never be included in JavaScript), plus JSX is compiled to hyperscript anyways (`React.createElement` is actually `h(tagOrComponent, props, children)`) so it's more apparent why it works how it does, plus we are already using functions as the building block for the rest of our code.

server.js
ahdinosaur: my reasoning is that this is not actually a special case, this is just a normal service. my intent in changing services to be the feathers plugin pattern, rather than a feathers service, was to avoid creating special cases, as i realized these are super common and i found it to become an anti-pattern to try and figure out the differences, especially when they all blend together (this is not the only service with hooks, for example).

ahdinosaur (approved):
whoop whoop! 💯 

left some comments, nothing major.

ahdinosaur:
oh also, can we remove https://github.com/root-systems/cobuy/pull/140/commits/f071cf60d65bd7e6cc1934dca80984c6dec0bbae from this pull request before we merge?

to do this, i'd to `git rebase -i origin/master` and remove the line for that commit.

sarah-arrrgh:
updated with suggested changes

sarah-arrrgh:
closes #32

PR #141, opened 2017-08-20
update product (was offering) component to match data model

### LINKED ISSUE #114: change offering form to match data model
Child task of #36

@ahdinosaur and i note certain structural differences between the data coming out of the offerings form, and the data model for resource types.

example of current offering form output ("one crate of 16 cartons of 12 eggs"):

```
offering: {
  name: "a crate of egg cartons",
  quantity: 1,
  unit: 'ea',
  resource: 'crate',
  offering: {
    quantity: 16,
    unit: 'ea',
    resource: 'carton',
    offering: {
      quantity: 12,
      unit: 'ea',
      resource: 'egg'
    }
  }
}
```

sample example given our data model:

```
resourceType: {
  name: "a crate of egg cartons",
  items: [ { value: 1, unit: "ea", resource: "crate" },
          { value: 16, unit: "ea", resource: "carton of eggs" } ]
}

resourceType: {
  name: "a carton of eggs",
  items: [ { value: 1, unit: "ea", resource: "carton" },
            { value: 12, unit : "ea", resource: "egg" } ]
}
```

notes on the comparison:

- the current output from the form privileges 'semantics', i.e. being more 'human-comprehensible'
- the data model privileges and presupposes a normalized, atomistic approach, where items/resources (especially the 'containers') can be reused

challenges for the form:

- can we get the best of both worlds? could we potentially do some normalization under the hood, while the UI is preserved?
- is there a limit to how much we could do under the hood? what if multiple different containers had the same name? how would we automate their reusability?

ahdinosaur:
fixes https://github.com/root-systems/cobuy/issues/114


ahdinosaur:
screenshot_20170821_120313

ahdinosaur:
screenshot_20170824_175755

resources/components/ResourceTypeEditor.js
```diff
@@ -0,0 +1,253 @@
...
+  return h('div', {
+    className: styles.itemListContainer
+  }, [
+    fields.map((field, index) => (
+      h(ResourceTypeItem, {
+        key: index,
+        styles,
+        field,
+        resourceTypes,
+        removeField: () => fields.remove(index)
```
iainkirkpatrick, line 133: this doesn't look like it's working - trying to remove parts in storybook does nothing :(

iainkirkpatrick (changes requested):
@ahdinosaur check the comment - also is wiring up this workflow part of another ticket? looks sweet otherwise :)

resources/components/ResourceTypeEditor.js
ahdinosaur: ah yeah, renamed from `removeItem` but didn't do below.

iainkirkpatrick (approved):
rad :) 👍

PR #143, opened 2017-08-21
fix sqlite3 version to 3.1.8 to avoid other module install errors

iainkirkpatrick:
easy


ahdinosaur:
solution from https://github.com/mapbox/node-sqlite3/issues/865#issuecomment-322779063

ahdinosaur:
chur!

PR #167, opened 2017-09-04
added hook to find current user, this id is then added if current buy…

Julietbc:
…ing group does not have admin

### DIFF
--- a/ordering/services/orders.js
+++ b/ordering/services/orders.js
@@ -45,9 +45,9 @@ function getCurrentUser (hook) {
 function groupHasNoAdminRelation(hook) {
   const relationships = hook.app.service('relationships')
   const groupId = hook.data.consumerAgentId
-  return relationships.find({ query: {targetId: groupId} }).then((relationship)=>{
-    return isEmpty(relationship)
-  })
+  return relationships
+    .find({ query: {targetId: groupId} })
+    .then(isEmpty)
 }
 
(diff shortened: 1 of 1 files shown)


ordering/services/orders.js
```diff
@@ -28,6 +32,38 @@ const hooks = {
...
+ .then(agent => {
+   hook.params.agent = agent
+ })
+ .then(() => hook)
+}
+
+function groupHasNoAdminRelation(hook) {
+  const relationships = hook.app.service('relationships')
+  const groupId = hook.data.consumerAgentId
+  return relationships.find({ query: {targetId: groupId} }).then((relationship)=>{
```
ahdinosaur: could be fancy and pass in `isEmpty` directly to `.then()`, otherwise it's `relationships` yeah?

ordering/services/orders.js
```diff
@@ -28,6 +32,38 @@ const hooks = {
...
+    return isEmpty(relationship)
+  })
+}
+
+function createGroupAdminRelation(hook) {
+  const relationships = hook.app.service('relationships')
+  const groupId = hook.data.consumerAgentId
+  const userId =  hook.params.agent.id
+  console.log(userId, 'the user id')
```
ahdinosaur: log log log

ahdinosaur (approved):
rad! 💖

PR #169, opened 2017-09-05
style: use pentatonic font size scale

### LINKED ISSUE #68: setup design style guide
when i want to work on a Cobuy component, it should be clear how to follow the Cobuy style.

this guide will operate as a source of truth for design decisions that affect the whole app, and should be updated after the initial creation to keep everything up to date as design decisions change.

maybe pull relevant content from material ui style guide:

- colors
- typography
- material ui theme

use `rem`!

maybe use typography js?

MAYBE LATER

- layout guide (overview of grid system)

ahdinosaur:
closes #68 

https://github.com/root-systems/cobuy/issues/68#issuecomment-322102143

### DIFF
--- a/app/themes/base.js
+++ b/app/themes/base.js
@@ -50,33 +50,35 @@ export default {
     // http://spencermortensen.com/articles/typographic-scale/
     // and
     // https://blog.madewithenvy.com/responsive-typographic-scales-in-css-b9f60431d1c4
-    '-5': '0.50000em',
+    // 
+    // px and pt values are based on a 1 em parent element with 16px (12pt) root size.
+    // http://pxtoem.com/
+    '-5': '0.50000em', // 8px, 6pt
     '-4': '0.57435em', 
     '-3': '0.65975em',
     '-2': '0.75785em',
     '-1': '0.87055em',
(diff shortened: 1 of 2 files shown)


app/themes/base.js
```diff
@@ -46,25 +45,42 @@ export default {
...
-    '5rem', // [10] 60pt
-    '6rem', // [11] 72pt
-    '8rem' // [12] 96pt
-  ],
+    '-5': '0.50000em',
+    '-4': '0.57435em', 
+    '-3': '0.65975em',
+    '-2': '0.75785em',
+    '-1': '0.87055em',
+    0: '1.00000em',
```
sarah-arrrgh: @ahdinosaur i'd find it really helpful to retain some of the info that was there previously showing how rem/em font sizes correspond to regulat pt font sizes, can you add that in please?

ahdinosaur: okay, i'll try to add direct conversion values from http://pxtoem.com/. i'm worried they will be misleading though, since an em value is multiplied by the parent em values, so if we do this properly (as described in the article which describes the pentatonic scale we are using) then the em values will almost never match the direct conversion values.

PR #173, opened 2017-09-06
[WIP] Feature/start order assign taskplans

### LINKED ISSUE #41: event: "order open!"
job story: when as an admin i've finished the order pre-requisites and i'm ready for my group to begin an order, i should be able to "open" the order. this should assign everyone in the group a TaskPlan to cast their individual intent, which should then send them an email notification to tell them of their current task.

sub-tasks:

- https://github.com/root-systems/cobuy/issues/161

iainkirkpatrick:
actually open / start an order

moves a bunch of the pre-req and relationship hooks onto the creation of the person agent, rather than when an order starts

closes #41


iainkirkpatrick:
superseeded by #192

PR #177, opened 2017-09-11
Feature/individual intent component

### LINKED ISSUE #49: component: individual intent casting

sarah-arrrgh:
closes #49


ordering/components/ProductList.js
```diff
@@ -0,0 +1,26 @@
...
+import ListViewProduct from './ListViewProduct'
+import styles from '../styles/ProductList'
+
+function ProductList (props) {
+  const { styles, products } = props
+
+  return (
+    h('div', {
+      className: styles.container
+    },
```
ahdinosaur: i think even though `map` returns an array, this should all be contained within an `[]` (see [`react-hyperscript` source](https://github.com/mlmorg/react-hyperscript/blob/master/index.js) to understand what is happening). then returned react elements from `map` need `key: product.id`.

app/locales/en.json
```diff
@@ -57,9 +57,10 @@
   "quantity.unit": "quantity unit",
   "currency.nzd": "NZD (New Zealand Dollars)",
   "ordering.minimum": "minimum",
-  "ordering.desired": "desired",
-  "ordering.maximum": "maximum",
+  "ordering.from": "from",
+  "ordering.add": "add",
   "ordering.intentTitle": "{name} per {step, number} {unit}",
+  "ordering.priceSpec": "if everyone orders more than {minimum}, then this item will cost ${price} each, in which case, I want to order ",
```
ahdinosaur: sweet!

ordering/components/ListViewProduct.js
```diff
@@ -0,0 +1,54 @@
...
+            className: styles.fromText
+          }),
+          ` $${product.priceSpecifications[0].price}`
+        ])
+      ])
+    ])
+  )
+}
+
+export default pipe(
```
ahdinosaur: can we use `compose` instead of `pipe`? i'd like us to be consistent, they have opposite orderings. i started using `pipe`, but have since realized that `compose` makes more sense (also is provided by `recompose`).

ordering/components/SingleViewProduct.js
```diff
@@ -0,0 +1,74 @@
...
+            className: styles.productText
+          }, product.description),
+          h('p', {
+            className: styles.priceText
+          }, [
+            h(FormattedMessage, {
+              id: 'ordering.from',
+              className: styles.fromText
+            }),
+            ` $${product.priceSpecifications[0].price}`
```
ahdinosaur: i reckon this should be part of the above internationalization text, so the message is like `ordering.price` and is passed in the price, like `from ${price}`, or something along those lines.

ordering/components/SingleViewProduct.js
```diff
@@ -0,0 +1,74 @@
...
+          ])
+        ]),
+        h('div', {
+          className: styles.priceSpecsContainer
+        }, map((priceSpec) => {
```
ahdinosaur: see previous comment about `map` and hyperscript.

ahdinosaur: also, the more "ramda-y" way to do this would be:

```js
const renderPriceSpecs = map(priceSpec => (
  h(ProductPriceSpec, {
    priceSpec,
    key: priceSpec.id
  })
))

    // ...
    }, [
      renderPriceSpecs(product.priceSpecs)
    ])
    // ...
```

ordering/stories/ProductList.js
```diff
@@ -0,0 +1,68 @@
...
+
+import ProductList from '../components/ProductList'
+
+// https://github.com/root-systems/cobuy/wiki/Models
+const mockProductList = [
+  {
+    name: 'crayons',
+    description: 'these are crayons',
+    image: 'http://www.mercurius-australia.com/site/images/1250623.jpg',
+    priceSpecifications: [
```
ahdinosaur: i changed this to `priceSpecs` (because it felt like soooooooo long to type), but happy to be convinced to change it back (since my other half of my self thinks sortened names are lazy). but we'll need to follow suit in the product editor code, which at the moment is using `priceSpec`.

ordering/stories/ProductList.js
```diff
@@ -0,0 +1,68 @@
+import { storiesOf } from '@storybook/react'
+import { action } from '@storybook/addon-actions'
+import { reduxForm, Field } from 'redux-form'
+import h from 'react-hyperscript'
+
+import ProductList from '../components/ProductList'
+
+// https://github.com/root-systems/cobuy/wiki/Models
+const mockProductList = [
```
ahdinosaur, line 7: yay sweet examples! 😊

ordering/components/ProductPriceSpec.js
```diff
@@ -0,0 +1,43 @@
...
+  const { styles, priceSpec } = props
+
+  return (
+    h('div', {
+      className: styles.container
+    }, [
+      h(FormattedMessage, {
+        id: 'ordering.priceSpec',
+        className: styles.priceSpecText,
+        values: {
```
ahdinosaur, line 21: this is rad, but shouldn't we pass in `priceSpec.currency` and use that as well in the internationalization message (instead of `$`, which only applies to currencies that use name dollar)? if we want to use a symbol, keen to have a hard-coded object that maps currencies to symbols, like https://github.com/root-systems/cobuy/blob/master/quantity/data/units.js, then use that.

ordering/components/SingleViewProduct.js
```diff
@@ -0,0 +1,74 @@
...
+            id: 'ordering.add',
+            className: styles.addButtonText
+          })
+        })
+      ])
+    ])
+  )
+}
+
+const ConnectedSingleViewProduct = reduxForm({ form: 'singleViewProduct' })(SingleViewProduct)
```
ahdinosaur: hey, the purpose of the `compose` / `pipe` chain for higher-order components is to avoid this. would be much better after `connectFela`. 😄

ordering/stories/ProductPriceSpec.js
```diff
@@ -0,0 +1,35 @@
...
+        component: ProductPriceSpec,
+        priceSpec: mockPriceSpec
+      })
+    ])
+  )
+}
+
+const ConnectedPriceSpecForm = reduxForm({ form: 'priceSpec' })(PriceSpecForm)
+
+storiesOf('ordering.ProductPriceSpec', module)
```
ahdinosaur, line 34: good work on the story (which includes this boilerplate)!

ordering/components/ProductFacet.js
```diff
@@ -0,0 +1,47 @@
...
+        className: styles.valuesContainer
+      }, [
+        h(Field, {
+          name: facet.name,
+          component: RadioButtonGroup,
+          className: styles.buttonsContainer
+        }, map((facetValue) => {
+          return h(RadioButton, {
+            label: facetValue.name,
+            value: facetValue.name
```
ahdinosaur: value would be `facetValue.id` right?

ahdinosaur (commented):
very legit, nice work! ✨ 

added some feedback 💚

ahdinosaur:
@sarah-arrrgh @iainkirkpatrick can we haz a screenshot or gif of what these components look like? keen to capture progress as we go, plus it communicates what this change does without having to spin up the app.

ordering/components/ProductPriceSpec.js
sarah-arrrgh: have changed it so it will show as `NZD7.99` instead of `$7.99`

ordering/components/SingleViewProduct.js
sarah-arrrgh: have changed it so it will show as `NZD7.99` instead of `$7.99`

ordering/components/ListViewProduct.js
sarah-arrrgh: done!

ordering/components/ProductFacet.js
sarah-arrrgh: done!

ordering/components/ProductList.js
sarah-arrrgh: done!

ordering/components/SingleViewProduct.js
sarah-arrrgh: done!

sarah-arrrgh:
@ahdinosaur ready for your perusal 🐱

PR #180, opened 2017-09-14
move authentication components to dogstack-agents

ahdinosaur:
corresponds with https://github.com/root-systems/dogstack-agents/pull/9

the reason is because i want to use these components in the dogstack example.


iainkirkpatrick (approved):
❤️  it! a question of whether we should merge this while it relies on a branch as a dependency? (dogstack-agents)

PR #184, opened 2017-09-22
tokens!

### LINKED ISSUE #159: token / capability service
we will need a few [token](https://en.wikipedia.org/wiki/Security_token) / [capability](https://en.wikipedia.org/wiki/Capability-based_security) features in Cobuy:

- [group invites](https://github.com/root-systems/cobuy/issues/93)
- [reset password](https://github.com/root-systems/cobuy/issues/58)
- secure download link

these are the services i've made so far with Mischa in Minded, but for Cobuy i wonder if it is advantageous to combine these into a single service.

possible data model:

```yml
- id (increments)
- agentId (integer)
- type (string)
- params (json)
- isConsumed (boolean)
- isRemoved (boolean)
```

where `type` is a sting referencing type of token: `inviteMember`, `resetPassword`, `downloadBlob`
where `params` would be flexible for each type of token:

group invites: `groupId`, `roles`
download: `blobId`

thoughts?

ahdinosaur:
closes #159

### DIFF
--- a/app/schemas.js
+++ b/app/schemas.js
@@ -3,9 +3,14 @@ const ajv = new Ajv()
 
 const taskPlanSchema = require('../tasks/schemas/taskPlan')
 const taskRecipeSchema = require('../tasks/schemas/taskRecipe')
+const tokenSchema = require('../tokens/schemas/token')
+const tokenConsumeSchema = require('../tokens/schemas/tokenConsume')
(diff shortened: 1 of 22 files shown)


iainkirkpatrick (approved):
read through, looks cool, can discuss in more detail in person :)

Issue #39, opened 2017-06-19
page: individual intent casting


ahdinosaur:
depends on https://github.com/root-systems/cobuy/issues/44 and https://github.com/root-systems/cobuy/issues/49.

job story: when my group is ready to begin an order, i should be assigned a TaskPlan to cast my individual intent, where i can say my { desired, minimum, maximum } values for how much i want to order for each product offered by the order supplier given the price specifications. 

branch: `feature/page-individual-intent-casting`

sarah-arrrgh:
TODO:

- [x] wire up the component to a page with a route
- [x] on dash, add another active task `cast your intent for order x` after an order has started
- [x] figure out how complete the intent casting component is

sarah-arrrgh:
**Services and hooks**

Compared with FA, we didn't develop the backend, we just used API routes that Sebastian built for us to interact with the database.

On Cobuy, Dogstack uses Feathers to generate and manage our API.
Feathers - service. Service refers to a collection of API routes for doing CRUD of a topic*, so we end up having an agent service and a order service etc.
e.g. if we want to create an order, we use the order service, and call .create on it if we want to 

**[Hooks](https://docs.feathersjs.com/api/hooks.html)**
If you have a basic service, you can attach hooks to that service. Hooks = functions that run either before a service attempts an operation (CRUD), or after it's successful, or if it errors.

---

*topic = a thing, like an order or an agent. FA = domain

CRUD = CREATE, READ, UPDATE, DELETE
Feathers READ = FIND or GET
Feather DELETE = REMOVE
https://docs.feathersjs.com/api/services.html
```
const myService = {
  find(params) {},
  get(id, params) {},
  create(data, params) {},
  update(id, data, params) {},
  patch(id, data, params) {},
  remove(id, params) {},
  setup(app, path) {}
}

app.use('/my-service', myService);
```

sarah-arrrgh:
syntax for reduxForm / higher-order components:

```
const higherOrderComponent = (component) => {
  const nextComponent = (props) => {
    return h(component, props)
  }
  return nextComponent
}
```

sarah-arrrgh:
examples:

```
function doubleFunction (fn) {
  return () => {
    return fn() * 2
  }
}

const four = () => 4
const eight = doubleFunction(four)
const eight = () => doubleFunction(four)()
```

sarah-arrrgh:
Status report:

[...]

PR #192, opened 2017-09-26
Feature/start order

### LINKED ISSUE #41: event: "order open!"

iainkirkpatrick:
- Wires up the order pre-reqs process under a parent Task of starting / completing an order
  - the parent Task is created when the user clicks the 'Start Order' button on the dashboard: a check is made with the database as to whether they still need to complete the pre-reqs or not
- Also under the parent Task is selecting a group, a supplier, and an admin for the order
- Once the parent Task has been completed (i.e. a respective TaskWork has been created) then created a `castIntent` TaskPlan for all members of the group.
- The `castIntent` TaskPlan isn't wired to any components yet, this is part of #39 (which can be implemented once this PR is merged)

closes #41


agents/components/SelectAgentForOrder.js
```diff
@@ -0,0 +1,72 @@
...
+      h(FormattedMessage, {
+        id: 'agents.profile',
+        className: styles.labelText
+      })
+    ]),
+    h('div', {
+      className: styles.innerContainer
+    }, [
+      h(Field, {
+        name: 'Select Group',
```
ahdinosaur: `name` does not mean `label` here, it means "path to save corresponding form value", as per how `name` works on an html `<input />`

agents/components/SelectAgentForOrder.js
```diff
@@ -0,0 +1,72 @@
...
+        floatingLabelText: 'Group',
```
ahdinosaur: needs i18n

agents/components/SelectAgentForOrder.js
```diff
@@ -0,0 +1,72 @@
...
+          })
+        }, agentCollection)
+      ]),
+      h(RaisedButton, {
+        className: styles.button,
+        type: 'button',
+        onClick: (ev) => {
+          // GK: not entirely clear why this is necessary considering the button type, but preventing default anyway
+          ev.preventDefault()
+          console.log('going to create a new agent / relationship')
```
ahdinosaur: does this do anything?

agents/getters/getCurrentAgentGroupIds.js
```diff
@@ -0,0 +1,19 @@
...
+import getRelationships from 'dogstack-agents/relationships/getters/getRelationships'
+import getCurrentAgent from 'dogstack-agents/agents/getters/getCurrentAgent'
+
+const getCurrentAgentGroupIds = createSelector(
+  getRelationships,
+  getCurrentAgent,
+  (relationships, currentAgent) => {
+    const getGroupIds = map(prop('targetId'))
+    const groupRelationships = getGroupIds(filter((relationship) => {
+      return relationship.sourceId === currentAgent.id // TODO: IK: possibly also specify relationshipType is admin or member
```
ahdinosaur: yeah filtering by any relationship with you as source will lead to future bugs. we'll need to filter by `member` relationships only.

i reckon it would be great to implement the reverse of [`getMembersByGroup`](https://github.com/root-systems/dogstack-agents/blob/master/agents/getters/getMembersByGroup.js), which will probably require the reverse of dependency getters.

agents/getters/getCurrentAgentGroupMemberIds.js
```diff
@@ -0,0 +1,19 @@
+import { createSelector } from 'reselect'
+import { filter, map, prop, values, contains } from 'ramda'
+
+import getRelationships from 'dogstack-agents/relationships/getters/getRelationships'
+import getCurrentAgentGroupIds from './getCurrentAgentGroupIds'
+
+const getCurrentAgentGroupMemberIds = createSelector(
```
ahdinosaur, line 7: i recommend using [`getMembersByGroup`](https://github.com/root-systems/dogstack-agents/blob/master/agents/getters/getMembersByGroup.js), this is missing some necessary logic in how to find group members.

agents/getters/getCurrentAgentGroupSupplierIds.js
```diff
@@ -0,0 +1,19 @@
+import { createSelector } from 'reselect'
+import { filter, map, prop, values, contains } from 'ramda'
+
+import getRelationships from 'dogstack-agents/relationships/getters/getRelationships'
+import getCurrentAgentGroupIds from './getCurrentAgentGroupIds'
+
+const getCurrentAgentGroupSupplierIds = createSelector(
```
ahdinosaur, line 7: where is this filtering for only relationships of type supplier?

agents/getters/getCurrentAgentGroupMemberProfiles.js
```diff
@@ -0,0 +1,19 @@
+import { createSelector } from 'reselect'
+import { filter, map, prop, values, contains } from 'ramda'
+
+import getProfiles from 'dogstack-agents/profiles/getters/getProfiles'
+import getCurrentAgentGroupMemberIds from './getCurrentAgentGroupMemberIds'
+
+const getCurrentAgentGroupMemberProfiles = createSelector(
```
ahdinosaur, line 7: i recommend removing this and similar getters (getCurrentAgentGroupMemberProfiles, getCurrentAgentGroupProfiles, getCurrentAgentGroupProfiles, and getCurrentAgentGroupSupplierProfiles), instead creating getters which transforms agent id's from corresponding getters into agents from `getAgents`, which will already have profiles attached.

app/locales/en.json
```diff
@@ -65,10 +65,17 @@
   "tasks.cancelTask": "cancel",
   "tasks.recipes.finishPrereqs": "finish pre-requisites for ordering",
   "tasks.recipes.setupGroup": "setup group profile and members",
+  "tasks.recipes.startOrder": "Start order",
```
ahdinosaur: so far we've been trying to have the translation strings be lowercase, then capitalize using css `text-transform`. but it's up to you, we can always "fix" this later.

ordering/getters/getOrders.js
```diff
@@ -0,0 +1,3 @@
+const getOrders = (state) => state.orders
```
ahdinosaur, line 1:
```js
import { prop } from 'ramda'
export default prop('orders')
```

😉 

but hey this is cool, just in case you were interested.

ordering/services/orders.js
```diff
@@ -136,3 +136,68 @@ function createSupplierAgent (hook) {
 function hasNoSupplierAgent (hook) {
   return isNil(hook.data.supplierAgentId)
 }
+
+function hasNotCompletedGroupOrSupplierProfile (hook) {
```
ahdinosaur, line 162: woop woop! 🎉

ordering/services/orders.js
```diff
@@ -136,3 +136,68 @@ function createSupplierAgent (hook) {
...
+      const supplierIds = map((relationship) => { return relationship.sourceId }, supplierRelationships)
+      return Promise.all([
+        profilesService.find({ query: { agentId: { $in: groupIds }}}),
+        profilesService.find({ query: { agentId: { $in: supplierIds }}})
+      ])
+    })
+  })
+  .then(profilesArray => {
+    // TODO: IK: need to return true if either the 'find' for group profiles with a name OR supplier profiles with a name returns true
+    return any((result) => { isNil(result) }, map((profiles) => { find((profile) => { !isNil(profile.name) }) }, profilesArray))
```
ahdinosaur: doesn't `(result) => { isNil(result) }` return nothing? i reckon you can just pass in `isNil`.

ahdinosaur: same for `(profile) => { !isNil(profile.name)`, could be `pipe(prop('name'), isNil, not)`.

ahdinosaur: sooo, does this work?

ahdinosaur: also you can remove `profilesArray` as the last argument to `any` and just pass the result directly in to `.then(...)`. point-free programming yay!

tasks/containers/StartOrderTask.js
```diff
@@ -0,0 +1,159 @@
...
+          service: 'profiles',
+          params: {
+            query: {
+              agentId: {
+                $in: currentAgentGroupIds
+              }
+            }
+          }
+        })
+        // get suppliers with a relationship to any groups of the currentAgent
```
ahdinosaur, line 65: should this query for supplier relationship?

tasks/containers/StartOrderTask.js
```diff
@@ -0,0 +1,159 @@
...
+          service: 'relationships',
+          params: {
+            query: {
+              sourceId: {
+                $in: currentAgentGroupIds
+              }
+            }
+          }
+        })
+        // get members with a relationship to any groups of the currentAgent
```
ahdinosaur: should this query for member relationships?

tasks/containers/StartOrderTask.js
```diff
@@ -0,0 +1,159 @@
...
+    actions: {
+      orders,
+      agents,
+      relationships,
+      profiles
+    },
+    query: (props) => {
+      var queries = []
+      const { taskPlan, selected } = props
+      const { currentAgent, currentAgentGroupIds, currentAgentGroupSupplierIds, currentAgentGroupMemberIds } = selected
```
ahdinosaur, line 33: nice

ahdinosaur (commented):
heyo, nice work! here's a splash of feedbacks 🌊

agents/getters/getCurrentAgentGroupMemberProfiles.js
iainkirkpatrick: hmmm - just checking in on this @ahdinosaur - as far as I can see, we will still need all these getters, otherwise the `shouldQueryAgain` function won't necessarily know whether to query the db or not?
i.e. a getter like `getCurrentAgentGroupSupplierProfiles` is used in that function to make sure we have the profiles of the suppliers of the groups of the currentAgent?
or maybe i'm not understanding what you mean by 'instead creating getters which transforms agent id's from corresponding getters into agents from getAgents'...

ahdinosaur: @iainkirkpatrick i'm not saying get rid of these getters completely, i'm saying i recommend having getters that use the `getAgents` getter to return _all the data for the same agents ids_. so instead of `getCurrentAgentGroupMemberProfiles`, it'd be `getCurrentAgentGroupMembers`, where each value is the full agent object with profile, credential, relationships, etc. does that make more sense?

ahdinosaur: something like

```js

const getCurrentAgentGroupMembers = createSelector(
  getAgents,
  getCurrentAgentGroupMemberIds,
  (agents, memberIds) => {
    return memberIds.map(memberId => agents[memberId])
  }
)
```

iainkirkpatrick: @ahdinosaur it does, but i'm still not sure how then we can manage the query loop - if we don't have some kind of check in `shouldQueryAgain` for whether the profiles of those members have loaded from the database, then the right condition in the `query` function won't ever be triggered to go and load the profiles?

maybe what i'm seeing here is that we are using getters for 2 purposes in `feathers-action-react` - for yanking data from state to give to wrapped components, but also to know whether we have the right data loaded yet to either stop querying or to activate new queries. I can see the point in not having `getCurrentAgentGroupMemberProfiles` for the first purpose, but don't see how we can avoid having it for the second? (could write conditional logic, but it'd be basically the same thing as the getter anyway?)

ahdinosaur: > but i'm still not sure how then we can manage the query loop - if we don't have some kind of check in shouldQueryAgain for whether the profiles of those members have loaded from the database, then the right condition in the query function won't ever be triggered to go and load the profiles?

you're right we still need these in `shouldQueryAgain`, i only mean we should get profiles from the agents in `getAgents`, so we have more re-usable getters.

ahdinosaur: it's some pattern like state -> state by relation -> enhanced data from composed state -> enhanced data by relation -> use case specific data

tasks/containers/StartOrderTask.js
iainkirkpatrick: done :)

tasks/containers/StartOrderTask.js
iainkirkpatrick: done :)

agents/components/SelectAgentForOrder.js
iainkirkpatrick: fixed :)

agents/components/SelectAgentForOrder.js
iainkirkpatrick: fixed :)

agents/components/SelectAgentForOrder.js
iainkirkpatrick: nope, button needs to be wired up to the create group / supplier / invite member flow later

agents/getters/getCurrentAgentGroupIds.js
iainkirkpatrick: done :)

agents/getters/getCurrentAgentGroupMemberIds.js
iainkirkpatrick: done :)

agents/getters/getCurrentAgentGroupSupplierIds.js
iainkirkpatrick: done :)

agents/getters/getCurrentAgentGroupMemberProfiles.js
iainkirkpatrick: done :)

app/locales/en.json
iainkirkpatrick: fixed :)

ordering/services/orders.js
iainkirkpatrick: not 100 what you meant by the last comment @ahdinosaur , but have updated with the other changes

iainkirkpatrick:
i'm gonna merge this for now @ahdinosaur, if there are other comments you (or anyone else) has i can update later :)

PR #194, opened 2017-10-05
added email, phone, website fields to profile

Julietbc:


agents/components/Profile.js
```diff
@@ -70,6 +70,48 @@ function Profile (props) {
...
+          name: 'email',
+          floatingLabelText: (
+            h(FormattedMessage, {
+              id: 'agents.email',
+              className: styles.labelText
+            })
+          ),
+          component: TextField,
+          fullWidth: true,
+          multiLine: true,
```
ahdinosaur: why `multiLine`?

ahdinosaur: i guess this question goes for all the new fields.

agents/schemas/profile.json
```diff
@@ -26,6 +26,18 @@
       "faker": "internet.avatar",
       "format": "uri",
       "description": "Agent avatar image link"
+    },
+    "email" : {
+      "type": "string",
```
ahdinosaur: we could add format email for this.

ahdinosaur (commented):
rubble dubble yub wub

iainkirkpatrick (approved):
think it's good for now - we can add the markdown editor later if we want, maybe more advanced schema stuff for fields as well. but doth demo waits for no human ;)

agents/schemas/profile.json
```diff
@@ -28,7 +28,7 @@
       "description": "Agent avatar image link"
     },
     "email" : {
-      "type": "string",
+      "type": "emails",
```
ahdinosaur: hey @iainkirkpatrick, i don't think this is legit JSON Schema.

i'd expect `"type": "string"`

and then another line `"format": "email"`.

or am i missing some special `emails` type?

iainkirkpatrick: ooooh my bad! good catch - just checked and i totally read the schema docs wrong! will fix with fix branch

PR #211, opened 2017-10-15
Feature/email invite group members

### LINKED ISSUE #161: email invite group members
depends on https://github.com/root-systems/cobuy/issues/159

email has basic content to say "hey, thanks for being a human on Cobuy!" with a token to create an account / login with existing account to be added to the group.

iainkirkpatrick:
Send an email to invited group members with a link to a page where they can set their password.
Update that user's credential with their new password on submission, then sign them in, then redirect them to `'/'`.

TODO:
- [x] redirect to home is currently breaking due to redux-form not destroying the form correctly, unsure why...

closes #161


iainkirkpatrick:
ooh, forgot to mention, this PR relies on [this one](https://github.com/root-systems/dogstack-agents/pull/16) in `dogstack-agents`

agents/services/agents.js
```diff
@@ -8,24 +8,40 @@ const hooks = {
...
           })
-          .then((cred) => {
+          .then((token) => {
             // TODO: IK: looks like the email isn't getting attached to the credential correctly, investigate dogstack-agents
             return hook.app.service('mailer').create({
               from: 'hello@cobuy.nz',
-              to: cred.email || 'no@email.com',
-              subject: 'Test Cobuy SignUp Email',
-              html: 'test email'
+              to: agentEmail || 'no@email.com',
```
ahdinosaur: this email should only be sent if you don't have an existing credential (password, oauth login, etc)

ahdinosaur: if you are the first one to create an account (the group admin), we need a separate email flow to verify their email

agents/components/Invited.js
```diff
@@ -0,0 +1,57 @@
...
+
+  return h('form', {
+    className: styles.container,
+    onSubmit: handleSubmit((data) => {
+      invited.invitedPatchPassword({ jwt, password: data.password })
+    })
+  }, [
+    h('h2', `You've been invited to Cobuy!`),
+    h('div', [
+      h('p', 'Enter a new password below'),
```
ahdinosaur, line 25: eventually we'll want to be able to use the oauth logins here. if not now, we should make a new ticket to do this later.

agents/components/Invited.js
```diff
@@ -0,0 +1,57 @@
...
+import styles from '../styles/Invited'
+
+function Invited (props) {
+  const { styles, handleSubmit, actions: { invited } } = props
+  const { jwt } = props.match.params
+
+  return h('form', {
+    className: styles.container,
+    onSubmit: handleSubmit((data) => {
+      invited.invitedPatchPassword({ jwt, password: data.password })
```
ahdinosaur, line 20: nice!

agents/dux/agents.js
```diff
@@ -0,0 +1,47 @@
+// TODO: IK: maybe rename this / move somewhere else
```
ahdinosaur: yeah, can we name this `dux/invited`, or something? i reckon it's confusing as agents.

agents/dux/agents.js
```diff
@@ -0,0 +1,47 @@
...
+// actions
+const payloadCtor = (cid, payload) => payload
+const metaCtor = (cid) => ({ cid })
+
+export const invitedPatchPassword = createAction('INVITED_PATCH_PASSWORD', payloadCtor, metaCtor)
+
+// epics
+export default combineEpics(patchPasswordAndSignIn)
+
+export function patchPasswordAndSignIn (action$, store, { feathers }) {
```
ahdinosaur: woop woop

agents/services/agents.js
```diff
@@ -8,24 +8,40 @@ const hooks = {
...
+              subject: `You're invited to join Cobuy!`,
+              html: `
+                Hi -name-. You've been invited by -admin- to join -group- on Cobuy! Click <a href=http://localhost:3000/invited/${token.jwt}>here</a> to set your password and start co-buying!
+              `
             })
           })
           .then(() => hook)
+          .catch((err) => console.log('error sending email invite,', err))
```
ahdinosaur: do we want to swallow this error?

agents/services/agents.js
```diff
@@ -8,24 +8,40 @@ const hooks = {
...
             // TODO: IK: looks like the email isn't getting attached to the credential correctly, investigate dogstack-agents
             return hook.app.service('mailer').create({
               from: 'hello@cobuy.nz',
-              to: cred.email || 'no@email.com',
-              subject: 'Test Cobuy SignUp Email',
-              html: 'test email'
+              to: agentEmail || 'no@email.com',
+              subject: `You're invited to join Cobuy!`,
+              html: `
+                Hi -name-. You've been invited by -admin- to join -group- on Cobuy! Click <a href=http://localhost:3000/invited/${token.jwt}>here</a> to set your password and start co-buying!
```
ahdinosaur: um, we can't hardcode the development url. keen for you to be able to access the base url at `app.get('url')`

tokens/services/tokenConsumes.js
```diff
@@ -61,11 +74,16 @@ function getPayload (hook) {
 }
 
 function callServiceMethod (hook) {
+  // TODO: IK: probably a better way to do this than serviceId, a permanent column in the token table?
```
ahdinosaur, line 77: yeah, keen for this refactor, thanks for the comment.

ahdinosaur (commented):
this is pretty legit, shot @iainkirkpatrick . :basketball: 

can merge now, but will need a couple changes before the start -> invite flow is usable in production.

agents/services/agents.js
ahdinosaur: depends on the relevant code from https://github.com/root-systems/cobuy/commit/fd64c4bfe53a71203a2730158b22123af9e09317

agents/services/agents.js
iainkirkpatrick: @ahdinosaur keen on your thoughts on this, looking at this problem again today...

so at the moment this is a after-create hook on `agents`, which gives us the problem that this hook fires before the `credentials` service has finished creating a users credential - meaning we can't tell if this is an invited user (through MemberInvites) or a user signing up by themselves.

so I thought aha! why not just a hook on `credentials` rather than `agents`? but after experimenting with that, i learned that currently the flow is:
- registering creates a `credential`, which then creates an `agent` if there is no agent data in the `before` hook
- creating that `agent` then creates a `credential` in the `after` hook (this credential has no data associated with it apart from `agentId`, i.e. no password or Oauth id)
so hooking onto `credentials` for email invites ends up sending 2 invites, which is bad, and would also send an invite for any other credential created for a user, also bad. 

i'm wondering now whether to prevent that second `credential` creation... but i can imagine situations where we want creating an `agent` to also create a `credential`. and that kinda strikes at the point as i'm thinking now - perhaps we should BE creating `agents` which then create `credentials`, rather than the other way around? `MemberInvites` works this way - is there a reason registering doesn't?

Issue #214, opened 2017-10-16
Order summary plumbing


michael-smith-nz:
depends on #210

Order Summary component and story has already been created but no real data has been used.

This ticket is to wire up the component to real data from feathers.

related to #178

Issue #191, opened 2017-09-26
Tax/GST


michael-smith-nz:
Are we needing to include tax/GST information in our orders?
For Tapin, I can imagine users needing to know tax exclusive and inclusive totals.

We could add the ability when creating a supplier to have the default items to be tax exclusive or inclusive, much like Xero do it. Then show this on the order summary.
![screenshot from 2017-09-26 10-21-18](https://user-images.githubusercontent.com/1152577/30852706-95ab6bbe-a2a4-11e7-8c23-850a71f00e38.png)

danalexilewis:
@NotThatSmith great point to raise.

My hunch would be this would be more consumer specific then supplier specific.

Ie a given agent/group/order could have the gst set to be included or excluded.

For example a business purchasing goods is more interested in the exclusive of gst cost while a food consumer is interested in the inclusive of gst cost.

I think people should be able to toggle it for themselves as best reflects their need. So it's probably part of the profile - maybe default on.

Issue #21, opened 2017-03-20
order status


ahdinosaur:
working on https://github.com/enspiral-root-systems/cobuy/issues/3 brought back memories of building [Cobudget](http://cobudget.co).

it'd be great to have a clear set of possible "status" that an order can be in.

the so far i made a `startDate` and `endDate` to delimit the initial "intent-casting" process, which makes the following statusus:

- pending
- open
- closed

but really i think there are more. what other modes can an order be in and what should we call all these possible states?

ahdinosaur:
some descriptions

1. while facilitator is editing order and before it becomes live
2. after facilitator clicks 'go live' and before individual intent phase completes
3. after individual intent phase completes and before group commitment is finalized
4. after group commitment is finalized and before money is paid
5. after money is paid and before invoice is sent to supplier
6. after invoice is sent to supplier and before it is acknowledged
7. after invoice is acknowledged and before shipment is received
8. after shipment is received and before shipment is accounted
9. after shipment is accounted and before shipment is divided
10. after shipment is divided and before shipment is packed
11. after shipment is packed and while packs are being picked up
12. after all packs are picked up

i probably missed some still... :smile:

ahdinosaur:
since this issue is now relevant again, i'd appreciate any feedback, then i reckon next step here is to give identifier names to these statuses / stages / phases / modes.

michael-smith-nz:
Nice @ahdinosaur! This is roughly how I see them looking from the users'(not admin) point of view:

Order doesn't exist yet/being drafted
- while facilitator is editing order and before it becomes live

Order open
- after facilitator clicks 'go live' and before individual intent phase completes

Order closed
- after individual intent phase completes and before group commitment is finalized

Requires Payment
- after group commitment is finalized and before money is paid
- after money is paid and before invoice is sent to supplier

Waiting on delivery
- after invoice is sent to supplier and before it is acknowledge
- after invoice is acknowledged and before shipment is received

Packing
- after shipment is received and before shipment is accounted
- after shipment is accounted and before shipment is divided
- after shipment is divided and before shipment is packed

Pick up
- after shipment is packed and while packs are being picked up

Order finished
- after all packs are picked up

danalexilewis:
@NotThatSmith & @ahdinosaur nice write ups.

I think the stages are slightly different depending on setup parameters - similar to the summary.

For example if purchases (schools) were buying individually then the payment stage would be order sent and maybe supplier has sent out invoices. 

[...]

PR #215, opened 2017-10-16
Feature/order summary

### LINKED ISSUE #178: component: order summary
this component displays information when an order is closed.

should include list of "commitments" given aggregate of intents.

TODO

- [ ] create sketch with component architecture / box model
- [ ] storybook each major component with various props

related to: #43, #40

michael-smith-nz:
This is a component and story wired up with dummy data for order summary.

Have created a plumbing ticket (#214) for when the Close an order ticket has been finished.

closes #178


ordering/containers/OrderSummary.js
```diff
@@ -0,0 +1,11 @@
+import h from 'react-hyperscript'
+import { compose } from 'recompose'
+
+// Only mock data at the stage, nothing in state or feathers.
+import getOrderSummaryProps from '../getters/getOrderSummaryProps'
+import OrderSummary from '../components/OrderSummary'
+
+export default compose(
```
iainkirkpatrick, line 8: my thinking is that we will have to connect to redux state in this container - as we access it from a new route, meaning it's possible no data is loaded yet. so IMO it'd look something like the other containers that use `feathers-action-react` do, with `query` and `shouldQueryAgain` functions to fetch all the data needed.

iainkirkpatrick (changes requested):
just the container i think could be upgraded, but open to whetehr this happens now or in a 'wiring' ticket

iainkirkpatrick:
think your opinion would be good here @ahdinosaur :)

ordering/containers/OrderSummary.js
michael-smith-nz: Yea, agreed. I did leave that out waiting to link with the feathers/database ticket. Happy to add that back, if that is possible without interacting with the database.

michael-smith-nz:
Hey @agentlewis @ahdinosaur and others. Keen to get feedback for on what else we need to include to show this off at the demo. Images below

Grouped by agent:

agent-order-summary

For supplier:

supplier-order-summary

danalexilewis:
@notthatsmith this is great!

In the context of the flow this will show the key points.

The only thing within the component we might want is gst.

Other then that there is other order stuff that may be in this or another component:

 - group name
 - group Contact details 
 - delivery address/es
 - dates - order placed, payment due by, expected delivery, order status (this is getting into the logistical side - we could just make an order status that the order admin can update when they get notified eg dispatched)

michael-smith-nz:
@agentlewis rad.

In terms of GST, I might quickly display this and leave the deep thinking to #191 as it could get quite in depth.

I can see the group Contact details and delivery address/es needed on the supplier order summary, but not agents summary

Like your idea of displaying the Order Status and not the dates. I see there's #21 on Order Status. @ahdinosaur do you know if we have something like this already?

danalexilewis:
@NotThatSmith - good stuff. Some more clarity:

There’s two use cases we need to cater for and may be different summary types 

 1. Where the supplier needs to send the goods to the individual schools. Ie delivery address on agent summary

 2. Where the supplier needs to send to one address. Ie delivery address on order summary.

We also need a similar arrangement for invoices.

 1. Details of on agent summary to invoice the individual schools

2. Details to invoice a representative on order summary

danalexilewis:
Oh and I think an indicative hard coded/mob field or something for gst is fine fornow

michael-smith-nz:
@agentlewis  interesting! So Schools will NOT be like cobuying groups; Schools will coordinate the order together, but will pay the supplier and be delivered goods separately?

danalexilewis:
@michael yes - well at least groups of schools buying computers will want to place the order together and pay individually and have the items shipped individually.

However a school using cobuy internally to coordinate departments may want to order together, have the accounts department pay and receive the items at the office.

So the four permutations we need to cover are:

 - 1 order & 1 delivery address 
 - many orders & 1 address
 - 1 order & many addresses
 - many orders & many addresses

In the fisrst case we really don’t need the agent orders to be sent through - in the other 3 we do.

Happy to schedule a call if you would like to talk this through?

michael-smith-nz:
@agentlewis oh cool, this is good to sort out and also getting quite complex. Let's just focus on Tapin for now, I might confuse us by comparing it with cobuying groups.

> 1 order & 1 delivery address

- As a buyer, I want to order directly from the supplier and get it delivered to me
- As a supplier, I want one order summary with one delivery address

> many orders & 1 address

- As a buyer, I want to order with people INSIDE my school and get it delivered to me
- As a supplier, I want one order summary with one delivery address

> 1 order & many addresses

-  As a buyer, I want to order with people OUTSIDE of my school and get it delivered to me
- As a supplier, I want one order summary with many delivery addresses 

> many orders & many addresses

- Is this not the same as '1 order & 1 delivery address' at scale?

danalexilewis:
@NotThatSmith Whenever I have writ order think person to send the invoice to.

So the many to many is:

Many invoices to be sent and many delivery addresses to be delivered to. Best example of this would be a national purchase of chrome books.

In regards to the same as 1 order and 1 address - i think this would be an example of a school buying something like stationary where they don’t want to internally coordinate what type and how much. Rather someone just buy some - oh another might be a single department like art that don’t need to order with anyone else. This is important for smaller orders.

An example of internal coordination would be furniture. 

Does that answer the question?

michael-smith-nz:
@agentlewis Cool, cool. That clears up the difference between a 1-1 and many-1.

The confusion is in the word 'Order' where what I think the word we should be using is Order Intent.

This might be getting too meta, but an Order is a combination of Order Intents. Using that wording I'll change the above to:


> 1 order, 1 order intent & 1 delivery address

- As a buyer, I want to order directly from the supplier and get it delivered to me
- As a supplier, I want one order summary with one delivery address

> 1 order, many order intents & 1 address

- As a buyer, I want to order with people INSIDE my school and get it delivered to me
- As a supplier, I want one order summary with one delivery address

> 1 order, many order intents & many addresses

-  As a buyer, I want to order with people OUTSIDE of my school and get it delivered to me
- As a supplier, I want one order summary with many delivery addresses 

> many orders, many order intents & many addresses

- Is this not the same as '1 order, many order intents & 1 address' at scale?

danalexilewis:
Yes if you add that in the last is redundant.

Wondering if there would ever be.

1 order, 1 order intent and many addresses?

Following the pattern...

Could be something like the ministry buying 1 of something for everyone.

danalexilewis:
@NotThatSmith I had a thought that maybe we should have some question checkboxes at the order setup phase:

Do order participants receive their invoice and pay Individually?
Do order participants have individual shipping addresses?

Then we can default them to checked and that way you will know what the summaries should look like.

ahdinosaur (approved):
this is great! 🐁 🧀 🍷

ahdinosaur:
ooh, this conversation thread is great. how and where can we save this information for later?

PR #216, opened 2017-10-17
dashboard only requests taskPlans and taskWorks of currentAgent

### LINKED ISSUE #213: Every user sees all users tasks
Currently it looks like there is no logic to load only the current user's tasks - all uncompleted tasks in the system are loaded into 'Your Active Tasks'.

iainkirkpatrick:
Previously when loading the dashboard, all `taskPlans` and `taskWorks` in the database would be loaded and displayed - even those that don't belong to the current user. Now only those that belong to the current user are loaded.

In the future we will want to protect the routes themselves against unauthorised data access.

closes #213

### DIFF
--- a/app/containers/Dashboard.js
+++ b/app/containers/Dashboard.js
@@ -39,6 +39,7 @@ export default connect({
   },
   shouldQueryAgain: (props, status) => {
     const { currentAgent, taskPlans } = props.selected
+    if (status.isPending) return false
     if (currentAgent && isEmpty(taskPlans)) return true
     return false
   }
--- a/app/containers/Dashboard.js
+++ b/app/containers/Dashboard.js
@@ -1,4 +1,5 @@
 import { connect } from 'feathers-action-react'
+import { isEmpty } from 'ramda'
 
 import Dashboard from '../components/Dashboard'
 import { actions as taskPlanActions } from '../../tasks/dux/plans'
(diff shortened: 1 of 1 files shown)


michael-smith-nz:
Hmmm there seems to be an infinite loop when loading the dashboard.

Logged in as `test@test.nz` after a `npm run burnthemall` and `npm run burnthedb`

Screenshot of logs, there's hundreds of these:
screenshot from 2017-10-17 11-04-44

I suspect you're requerying when `taskWorks` is empty, even when it is intentionally empty (for a brand new user)?

michael-smith-nz (changes requested):
comments above

iainkirkpatrick:
hmmm @NotThatSmith i'll check this out in the morning :)

iainkirkpatrick:
yay for manual testing tho!! :tada:

iainkirkpatrick:
ah so the problem is i didn't test this with a freshly burned db - infinite loop is because if there are no TaskPlans then `shouldQueryAgain` always returns true. will fix (somehow)

iainkirkpatrick:
actually that wasn't the problem - because at the moment a user always has an active task (though i can imagine in the future they might not, so this is something to think about). the problem was not guarding against querying again when the taskPlans request was in-flight (as per my understanding). fixed and re-pushed :)

ahdinosaur:
⛵️ 🌞

PR #217, opened 2017-10-17
each different use of Profile provides a different string for '{agent…

### LINKED ISSUE #206: profile editor says "My Profile", even when on group or supplier
this is confusing, here's a ticket to fix this

iainkirkpatrick:
…Type} Profile'

Fixes using the `Profile` component and being stuck with "My Profile" header text - now instead of "My" we have either "Group", "Supplier" or "My" depending on which profile you are filling out.

closes #206


michael-smith-nz (approved):
Niiiice

PR #218, opened 2017-10-17
Features/intent casting

Julietbc:


ahdinosaur:
TODO

- [x] get reproducible way to access page with real data
  - [x] snapshot (save) database in case you want to restore later
- [x] add queries for any and all relevant data for completing the task
  - [x] query for products
  - [x] query for resource types
  - [x] query for price specs
- [x] wire (copy from any other similar tasks, like profile update tasks) action creators that are used in task (creating or updating intents) to pass down as a function for onSubmit of intent cast forms
- [x] access routing params within task container, to selectively render either list view `/tasks/12` or single view `/tasks/12?productId=3` depending on route params
  - [x] create `getRouterSearchParams` to parse `state.routing.search` using `query-string` module.
  - [x] create `getCurrentProductId` to get current product id from params
  - [x] create `getCurrentProduct` to get current product from products and current product id
  - [ ] add this to current task container selector
  - [ ] create "router" component in task container render to selectively render either single view or list view depending on whether product id is present
- [ ] plug in components, giving them any props they need (see storybook for example usage)

Julietbc:
@NotThatSmith if you wanna jump in on this the next task is the routing .... if you don't get to it I will pair with Mikey on it tomorrow

ordering/components/ListViewProduct.js
```diff
@@ -1,17 +1,26 @@
...
 import { connect as connectFela } from 'react-fela'
 import h from 'react-hyperscript'
+import { Link } from 'react-router-dom'
+import FlatButton from 'material-ui/FlatButton'
 
 import { FormattedMessage } from '../../lib/Intl'
 
 import styles from '../styles/ListViewProduct'
 
+ const sortPriceSpecs = sortBy((priceSpec) => {
```
ahdinosaur, line 12: can be `sortBy(prop('price'))`

tasks/components/CastIntentTask.js
```diff
@@ -0,0 +1,74 @@
...
+        agentId: currentAgent.id,
+        desiredQuantity: parseInt(quantity),
+        productId: product.id,
+        priceSpecId: parseInt(split('-', priceSpecString)[1]),
+        orderId: taskPlan.params.orderId
+      }
+    }, value.priceSpecs))
+
+    forEach((submittedOrderIntent) => {
+      const scopedSubmittedOrderIntent = pick(['orderId', 'priceSpecId', 'productId'], submittedOrderIntent)
```
ahdinosaur, line 46: recommend splitting ramda calls into two steps:

1. create the function
2. use the function

so for example, this line could be

```js
const getScopedSubmittedOrderIntent = pick(['orderId', 'priceSpec', 'productId'])
const scopedSubmittedOrderIntent = getScopedSubmittedOrderIntent(submittedOrderIntent)
```

and in this case, the first line could be moved to the top-level scope of this file, it doesn't depend on any variables inside this current scope.

[why ramda?](http://fr.umio.us/why-ramda/)
[why curry helps](https://hughfdjackson.com/javascript/why-curry-helps/)

tasks/components/CastIntentTask.js
```diff
@@ -0,0 +1,74 @@
...
+      const existingOrderIntent = find((orderIntent) => {
+        const scopedOrderIntent = pick(['orderId', 'priceSpecId', 'productId'], orderIntent)
+        return equals(scopedOrderIntent, scopedSubmittedOrderIntent)
+      }, values(orderIntents))
+      console.log(orderIntents)
```
ahdinosaur, line 51: leftover `console.log`

tasks/components/CastIntentTask.js
```diff
@@ -0,0 +1,74 @@
...
+        desiredQuantity: parseInt(quantity),
+        productId: product.id,
+        priceSpecId: parseInt(split('-', priceSpecString)[1]),
+        orderId: taskPlan.params.orderId
+      }
+    }, value.priceSpecs))
+
+    forEach((submittedOrderIntent) => {
+      const scopedSubmittedOrderIntent = pick(['orderId', 'priceSpecId', 'productId'], submittedOrderIntent)
+      const existingOrderIntent = find((orderIntent) => {
```
ahdinosaur, line 47: this data massaging should be done upfront in getters.

tasks/containers/CastIntentTask.js
```diff
@@ -0,0 +1,110 @@
...
+          service: 'products',
+          params: {
+            query: {
+              supplierAgentId
+            }
+          }
+        })
+      }
+      if (!isEmpty(selected.products)) {
+        const resourceTypeIds = values(map((product) => {
```
ahdinosaur, line 46: could be

```js
const getResourceTypeIdsFromProducts = pipe(
  map(prop('resourceTypeId')),
  values
)
const resourceTypeIds = getResourceTypeIdsFromProducts(selected.products)
```

where first function can be defined in the top-level scope.

tasks/containers/CastIntentTask.js
```diff
@@ -0,0 +1,110 @@
...
+            query: {
+              id: {
+                $in: resourceTypeIds
+              }
+            }
+          }
+        })
+      }
+      if (!isEmpty(selected.products)) {
+        const productIds = values(map((product) => {
```
ahdinosaur, line 61: same as last comment

ahdinosaur (commented):
hey hey, left some feedback, i've already merged and am touching this code meow so now pressure to "fix" anything, just trying to share notes.

tasks/getters/getCastIntentTaskProps.js
```diff
@@ -0,0 +1,16 @@
...
+import getCurrentAgent from 'dogstack-agents/agents/getters/getCurrentAgent'
+import getProducts from '../../supply/getters/getProducts'
+import getResourceTypes from '../../resources/getters/getResourceTypes'
+import getPriceSpecs from '../../supply/getters/getPriceSpecsByProduct'
+import getRouterParams from '../../app/getters/getRouterParams'
+import getOrderIntents from '../../ordering/getters/getOrderIntents'
+
+export default createStructuredSelector({
+  currentAgent: getCurrentAgent,
+  products: getProducts,
```
ahdinosaur, line 11: oh, also. this getter returns *all the products* that have been loaded. but in the respective query and render code, it's being treated as if it's the products for the given order. in this case we want a getter for the current order and a getter indexing the products by order (creating an object where keys are order ids and values are arrays of products with the corresponding order id). happy to do this.

ahdinosaur: oops, products don't relate to an order directly, they relate to suppliers which relate to an order. soo, group products by supplier, key supplier by order, given order get supplier and then get products.

ahdinosaur: we already have `getCurrentSupplierProducts`. :sweat_smile:

tasks/components/CastIntentTask.js
```diff
@@ -0,0 +1,74 @@
...
+  const nextProps = merge(props, { product, onSubmit })
+  // TODO pass onNavigate to allow SingleViewProduct
+  // to navigate back to list product view
+  return h(SingleViewProduct, nextProps)
+  function onSubmit (value) {
+
+    const submittedOrderIntents = values(mapObjIndexed((quantity, priceSpecString) => {
+      return {
+        agentId: currentAgent.id,
+        desiredQuantity: parseInt(quantity),
```
ahdinosaur, line 38: by the way, this number should stay a string: https://github.com/root-systems/cobuy#numbers

PR #220, opened 2017-10-17
use a temp wrapped Switch to address react-router-redux bug until ups…

### LINKED ISSUE #196: clicking back changes the url but doesnt refresh content
tested in dev:

given I have opened localhost:3000
and I am not logged in
and I seed the splash page with sign in and register
and I click register/signup
and am taken to the corresponding form
When I click the browser back button
then the url changes
but not the content

browser: chrome
os: linux

iainkirkpatrick:
…tream fix

addresses not always being able to navigate back in history with the back button
see [this discussion](https://github.com/ReactTraining/react-router/issues/5072) for explanations.

this solution is temporary, to be replaced when `react-router-redux` sorts it out

closes #196

### DIFF
--- a/app/components/Layout.js
+++ b/app/components/Layout.js
@@ -1,8 +1,10 @@
 import h from 'react-hyperscript'
 import { compose } from 'recompose'
 import { connect as connectFela } from 'react-fela'
-import { Route, Switch } from 'react-router-dom'
+import { Route } from 'react-router-dom'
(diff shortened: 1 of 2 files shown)


ahdinosaur (approved):
shit yeah! this is great. :+1: 

@mischa-s and i found the same bug in MINDED, i forgot about this.

danalexilewis:
Should we put this in the dogstack example?

PR #243, opened 2017-10-20
wrap selectAgentForOrder form in order to pass down selected values

### LINKED ISSUE #239: show confirmed selection in dropdown when picking group, supplier, admin in order setup
Currently the selected option doesn't display, which results in bad UX

iainkirkpatrick:
fixes the selection forms for an order group, supplier, and admin not displaying the users selection once they've made it

NB. the loading for the selection task is quite slow, something to improve in future i think

closes #239

### DIFF
--- a/agents/components/SelectAgentForOrder.js
+++ b/agents/components/SelectAgentForOrder.js
@@ -25,7 +25,7 @@ const SelectAgentForOrderForm = compose(
       className: styles.innerContainer
     }, [
       h(Field, {
-        name: `select${agentType}`,
+        name: 'selectAgent',
         floatingLabelText: (
           h(FormattedMessage, {
             id: 'agents.labelText',
@@ -63,8 +63,7 @@ const SelectAgentForOrder = (props) => {
     form: `select${props.agentType}`,
     enableReinitialize: true,
     initialValues: {
-      // TODO: IK: is there a better way to write this
-      [`select${props.agentType}`]: props.selectedAgent ? props.selectedAgent.id : null
+      selectAgent: props.selectedAgent ? props.selectedAgent.id : null
     }
   })
   return h(SelectAgentForOrderForm, nextProps)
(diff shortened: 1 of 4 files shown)


agents/components/SelectAgentForOrder.js
```diff
@@ -53,13 +56,18 @@ function SelectAgentForOrder (props) {
...
   ])
+})
+
+const SelectAgentForOrder = (props) => {
+  const nextProps = merge(props, {
+    form: `select${props.agentType}`,
+    enableReinitialize: true,
+    initialValues: {
+      // TODO: IK: is there a better way to write this
+      [`select${props.agentType}`]: props.selectedAgent ? props.selectedAgent.id : null
```
ahdinosaur: why not call the field `selectAgent`?

agents/components/SelectAgentForOrder.js
```diff
@@ -53,13 +56,18 @@ function SelectAgentForOrder (props) {
       }, 'create new')
     ])
   ])
+})
+
+const SelectAgentForOrder = (props) => {
+  const nextProps = merge(props, {
+    form: `select${props.agentType}`,
```
ahdinosaur: `select-${props.agentType}`?

ahdinosaur (approved):
legit 🕴

agents/components/SelectAgentForOrder.js
iainkirkpatrick: is this for adherence to a common convention, or something else i'm missing? @ahdinosaur

agents/components/SelectAgentForOrder.js
iainkirkpatrick: hmm my understanding from other areas of the code was that this key needed to be named the same as the form... maybe not though? @ahdinosaur

iainkirkpatrick: oh sorry i think i see what you mean, yep can do

agents/components/SelectAgentForOrder.js
ahdinosaur: `selectsupplier` is hard to read as "select supplier". there's many possible conventions we could choose from: we could use camelCase, but to avoid capitalization pain and since it's common in DOM values in this case i've been using hypen-case

iainkirkpatrick: 👍

PR #249, opened 2017-10-24
add <title>Cobuy</title>

### LINKED ISSUE #244: add <title>Cobuy</title>
at the moment we have no `<title>` tag, so the page is named whatever the browser feels like.

this ticket is to add `<title>Cobuy</title>` in the layout.

can be done using:

- [`react-document-title`](https://github.com/gaearon/react-document-title)
- [`react-helmet`](https://github.com/nfl/react-helmet) (this will make it easy to add other head tags later, like facebook metadata)

ahdinosaur:
fixes https://github.com/root-systems/cobuy/issues/244

using `react-helmet`


iainkirkpatrick (approved):
looks rad to me :) 👍

PR #251, opened 2017-10-24
Feature/tapin demo seed data

### LINKED ISSUE #252: Seed data for Tapin demo
Add seed data that makes videoing a demo for Tapin presentations easier

iainkirkpatrick:
Adds seed data for the upcoming Tapin demo:

- group
- supplier to the group
- members, including a group admin, which belong to the group
- products (resourceTypes, products, and priceSpecs) that the supplier supplies

The idea is that we can record a demo video without needing to go through and invite new members, write a bunch of product data each time - we can just log in as the various members on browsers and demonstrate the ordering / intent casting functionality.

closes #252

### DIFF
--- a/db/seeds/tapin-demo.js
+++ b/db/seeds/tapin-demo.js
@@ -1,6 +1,17 @@
 const hasher = require('feathers-authentication-local/lib/utils/hash')
 const { flatten, map, pick, merge } = require('ramda')
 
+const groups = [
+  {
+    name: "Ruru Kāhui Ako",
+    description: "Ruru Kāhui Ako"
+  },
+  {
+    name: "School Supplies R'Us",
+    description: "If you're yearning for some learning, we'll set your mind churning"
+  }
+]
+
 const agents = [
   {
     name: 'Dan Lewis',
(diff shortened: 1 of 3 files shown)


db/seeds/tapin-demo.js
```diff
@@ -0,0 +1,208 @@
...
+      }))
+    })
+    // not sure why, but credentials .returning doesn't return the correct agentIds, hack for now
+    .then(() => ids)
+  })
+  .then((ids) => {
+    // insert person relationships
+    return Promise.all(agents.map((agent, i) => {
+      const relationship = {
+        relationshipType: i === 0 ? 'admin' : 'member',
```
sarah-arrrgh: @iainkirkpatrick - make dan a member

sarah-arrrgh (changes requested):

ahdinosaur (commented):
looks good.

agree with @sarah-arrrgh that all group admins (only dan in this case) _must_ also be group members, so probably makes sense to create all member relationships then any special relationships.

iainkirkpatrick:
fixed :)

PR #256, opened 2017-10-25
fix storybook

ahdinosaur:
this pull request fixes storybook. ~it also breaks the app, so don't merge.~

~but feel free to use and update this branch whenever you want to use storybook.~

~will need to investigate a proper fix. the problem is due to https://github.com/webpack/webpack/issues/3997 in `config/default.js`, which seems strange because it never uses `import`. it also can't use `import` and `export` because in the app it's loaded by [`node-config`](https://github.com/lorenwest/node-config) which doesn't understand ES modules.~

### DIFF
--- a/agents/dux/invited.js
+++ b/agents/dux/invited.js
@@ -3,7 +3,7 @@ import createAction from '@f/create-action'
 import { push } from 'react-router-redux'
 
 import { combineEpics } from 'redux-observable'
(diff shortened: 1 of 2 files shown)


ahdinosaur:
hey, i got this working! review pls?

iainkirkpatrick (approved):
keeeewl :)

PR #259, opened 2017-10-26
fix castIntentTask using the actual order data rather than params fro…

iainkirkpatrick:
…m taskPlan

a final fix for casting intents, don't use the `taskPlan` params for `supplierAgentId` as explained to @ahdinosaur


server.js
```diff
@@ -7,7 +7,7 @@ const services = [
   require('./ordering/services/orders'),
   require('./credentials/services/credentials'),
   require('./ordering/services/orderIntents'),
-  require('./agents/services/agents'),
+  require('./credentials/services/credentials'),
```
ahdinosaur: this happens two lines above?

PR #268, opened 2017-10-30
Close an order

### LINKED ISSUE #210: Close an order
A few possible ways of an order closing

- Time-based -  when an order end date is reached that order is considered closed.
- Manually close - An admin must manually press a button to close the order
- Something else?

In the future, we'll need to add some form of 'hooks' (send email, create tasks, etc) when an order is closed.

michael-smith-nz:
Closes #210


michael-smith-nz:
Current thoughts:

- Create a `closeOrder` task plan when the order is created.
- If order has an admin, assign the task plan to that admin
  - If no admin then assign to logged in user/user who started the order
- When admin is updated on order, also update on task plan (through hook)
  - Update only when task plan has not been completed (no task work exists for it)

More notes on #210 

Anyone can help pick this up if they want. A good place to start would be the empty functions I've created here - https://github.com/root-systems/cobuy/blob/feature/close-order/ordering/services/orders.js#L228-L242

michael-smith-nz:
Admin cannot be set via start an order screen, still investigating

iainkirkpatrick:
thoughts looking at this today:

- IMO we should create the 'closeOrder' task when the order is started, rather than when the order is created... doesn't make sense to me to close the order when it hasn't been started yet. this would involve swapping the current hooks that create the taskPlan to the `taskWork` service hooks, and listening to the same 'completeOrder' taskRecipes that are currently being listened to in order to create the 'castOrderIntent' taskPlans
- creating a taskWork for this `closeOrder` taskPlan (i.e. clicking the 'done' button) should create another taskWork for the associated order 'castIntent' taskPlans that are assigned to each member of the group - i.e. once the order is closed, noone can cast intents anymore.
- closing an order could also create a task to view an OrderSummary? if we don't do it this way (at least for now) then we'll need some other UI to access the order summary.

iainkirkpatrick:
end of day:

- closeOrder task is now created when the order is 'started', rather than created (as per first bullet above)
- a taskWork is created for each member of the group to close their `castIntent`taskPlan when the order closes (as per second bullet above), though note the user can still see the `castIntent` task in their list of tasks until #XXX is merged
- in addition to previous comment bullets, we also need to create `orderPlans` derived from each group member's `orderIntent`s (of those who did actually cast their intents) when the order closes. there is a TODO note on L39 of the taskWorks service file for this.

this isn't too far away! :) will hand-over with @NotThatSmith tonight

ahdinosaur:
> IMO we should create the 'closeOrder' task when the order is started, rather than when the order is created... doesn't make sense to me to close the order when it hasn't been started yet

hi, can we please be more specific than naming this "close order"? it's confusing because **the order is not closing**, we're still planning to buy the products from the supplier, we're only ending the currently active intent casting phase and moving in to the next phase of the order.

iainkirkpatrick:
hmm, i guess i used 'closeOrder' because that's what we have been using so far, feels useful to use the terms used in the code of this PR when describing the functionality of it. have you got an alternative suggestion for the name of this task @ahdinosaur?

ahdinosaur:
> feels useful to use the terms used in the code of this PR when describing the functionality of it

yeah, i agree our vocabulary should be consistent between code and context. my recommendation is to change the code to be "closeIntentPhase", or something similar that is better suited for describing this particular change in order status: #21

iainkirkpatrick:
todo next: 

- create a taskPlan to view an OrderSummary when an 'order closes'

IMO that's this ticket done after that, we can have another ticket for wiring up the OrderSummary with the correct data

iainkirkpatrick:
`ViewOrderSummary` dummy container / component added, and a `taskPlan` created to view em when an order 'closes'. IMO this ticket is ready for review (though i know some containers could use a clean on their queries, i might do that now) @NotThatSmith

iainkirkpatrick:
thought: should the 'closing order' page show the current state of the order? or is that just too much added complexity to have some summation of everyone's current `orderIntents`? are there any other ideas about what a manual 'close' of an order would look like? @NotThatSmith @agentlewis @ahdinosaur

iainkirkpatrick:
ready for review @ahdinosaur ? if no time then @NotThatSmith and I can discuss it and review :)

db/migrations/20171107113437_create-orderPlans.js
```diff
@@ -0,0 +1,16 @@
+exports.up = function(knex, Promise) {
+  return Promise.all([
+    knex.schema.createTableIfNotExists('orderPlans', function (table) {
+      table.increments('id')
+      table.integer('orderId').references('orders.id')
+      table.integer('agentId').references('agents.id')
+      table.integer('quantity')
```
ahdinosaur: all numbers should be saved as strings

tasks/components/CloseOrderTask.js
```diff
@@ -0,0 +1,14 @@
...
+
+export default (props) => {
+  const { taskPlan, currentOrder } = props
+  if (isNil(taskPlan)) return null
+  if (isNil(currentOrder)) return null
+  const { params: { orderId } } = taskPlan
+  const { consumerAgentId, supplierAgentId } = currentOrder
+
+  return h('div', {}, [
+    h('h2', {}, `Closing order ${orderId} for group ${consumerAgentId} from supplier ${supplierAgentId}`)
```
ahdinosaur, line 12: shouldn't this be internationalized?

tasks/components/ViewOrderSummaryTask.js
```diff
@@ -0,0 +1,7 @@
+import h from 'react-hyperscript'
+
+export default (props) => {
+  return h('div', {}, [
+    h('h2', {}, 'View Order Summary for ...')
```
ahdinosaur, line 5: more i18n

tasks/services/works.js
```diff
@@ -18,15 +18,58 @@ const hooks = {
...
-  || hook.data.taskRecipeId === 'completeOrderSetup'
+    || hook.data.taskRecipeId === 'completeOrderSetup'
+}
+
+function taskRecipeIsCloseOrder (hook) {
+  return hook.data.taskRecipeId === 'closeOrder'
+}
+
+function createOrderPlans (hook) {
+  // TODO: IK: this is half done, aiming towards creating an orderPlan for each user who created an orderIntent for an order when that order closes
```
ahdinosaur: if this is half done, what's not yet done?

tasks/services/works.js
```diff
@@ -36,26 +79,101 @@ function createCastOrderIntentTaskPlan (hook) {
...
+      const supplierAgentId = completeOrderTaskPlan.params.supplierAgentId
+      const params = { orderId, consumerAgentId: groupId, supplierAgentId }
+
+      return relationships.find({
+        query: {
+          sourceId: groupId,
+          relationshipType: 'member'
+        }
+      })
+        .then((relationships) => {
```
ahdinosaur: ramda golf

```js
.then(pipe(
  map(pipe(
    prop('targetId'),
    assigneeId => taskPlans.create({
      assigneeId,
      taskRecipeId,
      params
    })
  )),
  Promise.all
))
```

tasks/services/works.js
```diff
@@ -36,26 +79,101 @@ function createCastOrderIntentTaskPlan (hook) {
...
+          }))
+        })
+    })
+    .then(() => hook)
+}
+
+function createViewOrderSummaryTaskPlan (hook) {
+  const taskPlans = hook.app.service('taskPlans')
+  const taskRecipeId = taskRecipes.viewOrderSummary.id
+  // TODO: IK: at the moment the assigneeId is automatically the user who closes the order, should be the order admin
```
ahdinosaur: shouldn't the order admin be the only one is assigned the "close order" task?

ahdinosaur (commented):
looks good! still [concerned about using "close order" as our terminology](https://github.com/root-systems/cobuy/issues/21#issuecomment-342418946), but hey if others think this is best than okay.

tasks/components/CloseOrderTask.js
iainkirkpatrick: it could be, but it's only temporary text until we actually have a way to properly close an order... feels like it'd just be more code to remove later on.

tasks/components/ViewOrderSummaryTask.js
iainkirkpatrick: same as above

tasks/services/works.js
iainkirkpatrick: old comment, no longer applicable

tasks/services/works.js
iainkirkpatrick: oooh shiny :) needs to change a bit though now `[params, relationships]` are the args to that `.then` function...

tasks/services/works.js
iainkirkpatrick: fixed :)

PR #272, opened 2017-10-31
remove dogstack regrets

ahdinosaur:
what are these doing here?

some hangover from https://github.com/root-systems/dogstack/pull/77


iainkirkpatrick:
love the PR name 😸

PR #274, opened 2017-10-31
fix/product-list-querying

### LINKED ISSUE #269: problem with adding products for supplier post-reload
needs more investigation to pin down the specifics, but at the moment:

- when i start an order, i can setup a supplier and create a product
- if i don't enter any data for that product, when i reload, the product doesn't load and i can't create any additional products

iainkirkpatrick:
a fix for missing products in setupSupplier task by querying for taskPlan order in the setupSupplierTask container

had a thought to query for this at a 'higher' level, but wouldn't help in the Dashboard container, and otherwise it's just TaskWorker that has a container (and that seems too generic for specific queries like this)

closes #269


ahdinosaur (approved):
looks good to me!

PR #275, opened 2017-11-01
smush production bugs

ahdinosaur:
:boot: :bug: :boom:


ahdinosaur:
oops, reset db seed is not quite working in production, will fix. got an in-person review from @iainkirkpatrick so will merge when ready meow. :cat:

PR #276, opened 2017-11-01
fix(tasks): show only active taskPlans on dashboard

### LINKED ISSUE #228: bug: presence of a TaskWork doesn't prevent the associated TaskPlan from showing in active tasks
PR #276

sarah-arrrgh:
closes #228

### DIFF
--- a/tasks/getters/getActiveTaskPlans.js
+++ b/tasks/getters/getActiveTaskPlans.js
@@ -1,18 +1,11 @@
 import { createSelector } from 'reselect'
-import { map, prop, filter, contains, not, values } from 'ramda'
+import { prop, filter, pipe, isEmpty } from 'ramda'
 
-import getParentTaskPlans from './getParentTaskPlans'
-import getRawTaskWorks from './getRawTaskWorks'
+import getTaskPlans from './getTaskPlans'
 
 const getActiveTaskPlans = createSelector(
-  getParentTaskPlans,
-  getRawTaskWorks,
-  (taskPlans, taskWorks) => {
-    const taskWorksTaskPlanIds = map(prop('taskPlanId'), values(taskWorks))
-    return filter((taskPlan) => {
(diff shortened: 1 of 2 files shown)


tasks/getters/getActiveTaskPlans.js
```diff
@@ -1,13 +1,17 @@
...
 
 const getActiveTaskPlans = createSelector(
-  getTaskPlans,
-  (taskPlans) => {
-    const byDueDate = n => isNil(n.dueDate)
-    return filter(byDueDate, values(taskPlans))
+  getParentTaskPlans,
+  getRawTaskWorks,
+  (taskPlans, taskWorks) => {
+    const taskWorksTaskPlanIds = map(prop('taskPlanId'), values(taskWorks))
```
ahdinosaur: rather than doing the check for "isActive" in a separate getter, i think we should put this logic in the getter that enhances task plans, so for each task plan we add a derived property `isActive`. then if we want this getter, it's as easy as filtering by `isActive`. the difference is that with this current approach, if we are using `getTaskPlans` then we don't know which ones are active.

ahdinosaur: https://github.com/root-systems/cobuy/blob/master/tasks/getters/getEnhancedTaskPlans.js

ahdinosaur: hey, rather than using `ramda` as if it were `lodash` with the arguments in a strange reverse order, i recommend using `ramda` in two stages:

1. create a function without the data (this should be done at the highest scope possible, as in if it doesn't rely on any data variables then it can be done in the top-level scope next to where you have imports and such)
2. use the function on the data

so in this case, this code can be written as:

```js
// top-level scope
const getTaskPlanIdsFromTaskWorks = map(prop('taskPlanId'))

// in function scope (with data)
const taskPlanIds = getTaskPlanIdsFromTaskWorks(taskWorks)
```

PR #287, opened 2017-11-19
fix query logic to always retrieve relevant groups, suppliers, admins…

### LINKED ISSUE #281: bug: configuring order doesn't show available groups, suppliers, admins unless refreshed
as per title - to show the seeded data, for instance, requires a refresh

iainkirkpatrick:
… for order config

closes #281


ahdinosaur (approved):
shot 🏀

PR #288, opened 2017-11-19
fix assigning cast intent task to all group members when order closes

### LINKED ISSUE #285: task to 'add items to order' doesn't show for group members
it does appear for the order admin, but not for other members of the group

iainkirkpatrick:
closes #285

### DIFF
--- a/tasks/services/works.js
+++ b/tasks/services/works.js
@@ -35,7 +35,6 @@ function createCastOrderIntentTaskPlan (hook) {
   const relationships = hook.app.service('relationships')
   const taskRecipeId = taskRecipes.castIntent.id
   const completeOrderTaskPlanId = hook.data.taskPlanId
-  let params
 
   return taskPlans.get(completeOrderTaskPlanId)
   .then((completeOrderTaskPlan) => {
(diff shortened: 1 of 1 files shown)


ahdinosaur (approved):
shot 🏀

ahdinosaur:
oops, above review was meant for https://github.com/root-systems/cobuy/pull/287

tasks/services/works.js
```diff
@@ -31,31 +31,34 @@ function completeOrderSetupTaskWork (hook) {
 
 function createCastOrderIntentTaskPlan (hook) {
   const taskPlans = hook.app.service('taskPlans')
+  const orders = hook.app.service('orders')
   const relationships = hook.app.service('relationships')
   const taskRecipeId = taskRecipes.castIntent.id
   const completeOrderTaskPlanId = hook.data.taskPlanId
+  let params
```
ahdinosaur: what's up with this?

tasks/services/works.js
```diff
@@ -31,31 +31,34 @@ function completeOrderSetupTaskWork (hook) {
...
-    const groupId = completeOrderTaskPlan.params.consumerAgentId
     const orderId = completeOrderTaskPlan.params.orderId
-    const supplierAgentId = completeOrderTaskPlan.params.supplierAgentId
-    const params = { orderId, consumerAgentId: groupId, supplierAgentId }
-
+    return orders.get(orderId)
+  })
+  .then((order) => {
+    const { id, consumerAgentId, supplierAgentId, adminAgentId } = order
+    params = { orderId: id, consumerAgentId, supplierAgentId, adminAgentId }
```
ahdinosaur: why not pass this down the promise pipeline? it seems real sketchy (and prone to race conditions) to pass it through a global variable.

you could instead do:

```js
.then((order) => {
  const params = {}
  return Promise.all([
    Promise.resolve(params),
    relationships.find({})
  ])
})
.then(([params, relationships]) => {
})
```

ahdinosaur (changes requested):
i'm sketched out by the params variable. might be totally innocuous and legit, but concerned about future race conditions or spaghetti.

iainkirkpatrick:
good call, will change :)

PR #292, opened 2017-11-21
WIP: Chore/postgres local dev

### LINKED ISSUE #289: switch local dev to use postgres rather than sqlite
using sqlite for local dev has been easy, but now we have production / staging deployments which use postgres, we are encountering some bugs / crashes when deploying due to the differences between sqlite and postgres.

this ticket is to change connection config to postgres, along with a guide to getting cobuy running locally on MacOS / Linux.

iainkirkpatrick:
closes #289 

todo:
- [X] connection params to use postgres rather than sqlite
- [x] guide for macOS and Linux users

NB. currently running the app assumes you have a database in your local Postgres installation named `cobuy_development`. The guide will show users how to create this db, but could also be cool if possible to have an npm script (`npm run init`?) that does that, plus migrate, seed etc for them

### DIFF
--- a/tasks/services/works.js
+++ b/tasks/services/works.js
@@ -222,14 +222,17 @@ function createCastOrderIntentTaskWorks (hook) {
       // TODO: IK: feels like a pretty sub-optimal way of querying, probably makes sense to have orderId as it's own column
       return taskPlans.find({
         query: {
-          taskRecipeId: 'castIntent',
-          params: {
-            $like: `%"orderId":${orderId}%`
-          }
+          taskRecipeId: 'castIntent'
+          // TODO: IK: can't compare JSON values in postgres, need another way to do it
+          // params: {
+          //   $like: `%"orderId":${orderId}%`
+          // }
         }
       })
         .then((taskPlans) => {
-          return Promise.all(taskPlans.map((taskPlan) => {
+          // TODO: IK: can't compare JSON values in postgres, need another way to do it
+          const currentOrderTaskPlans = filter((plan) => { return plan.params.orderId === orderId }, taskPlans)
+          return Promise.all(currentOrderTaskPlans.map((taskPlan) => {
             const { id, taskRecipeId, assigneeId, params } = taskPlan
             return taskWorks.create({
               taskPlanId: id,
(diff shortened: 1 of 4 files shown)


iainkirkpatrick:
using the [dogstack guide for postgres setup](https://dogstack.js.org/guides/how-to-setup-sql-db.html) for now - for me on MacOS it really is as simple as installing postgres App, can someone confirm the process in that guide for Linux and then that Cobuy local dev works as expected with the new connection params? @ahdinosaur @NotThatSmith

michael-smith-nz:
@iainkirkpatrick  I've just added a password field for this to work, but the `burnthedb` script no longer works. I attempted to fix it myself, but my lack of postgres knowledge was showing. We can discuss at sprint planning

iainkirkpatrick:
@NotThatSmith interesting:

- was the password field required for connecting to the database on linux? i didn't seem to need one...
- when `burnthedb` fails, is that something in the migrations or seeds?

tasks/services/works.js
```diff
@@ -222,14 +222,17 @@ function createCastOrderIntentTaskWorks (hook) {
...
+          taskRecipeId: 'castIntent'
+          // TODO: IK: can't compare JSON values in postgres, need another way to do it
+          // params: {
+          //   $like: `%"orderId":${orderId}%`
+          // }
         }
       })
         .then((taskPlans) => {
-          return Promise.all(taskPlans.map((taskPlan) => {
+          // TODO: IK: can't compare JSON values in postgres, need another way to do it
```
ahdinosaur, line 233: https://www.postgresql.org/docs/current/static/functions-json.html

ahdinosaur:
> when `burnthedb` fails, is that something in the migrations or seeds?

the first thing done by `burnthedb` is remove the sqlite database file, so that part will no longer work.

fix incoming!

tasks/services/works.js
ahdinosaur: @iainkirkpatrick i did a thing! [`queryByOrder`](https://github.com/root-systems/cobuy/blob/a38b8f61b7255024d95fc7c5e364696e21cefe3c/tasks/hooks/queryByOrder.js)

iainkirkpatrick: oooh damn im intrigued to learn this wizardry @ahdinosaur 🎉 🎉 🎉

Issue #290, opened 2017-11-19
bug: don't need to set up group / supplier if already part of one


iainkirkpatrick:
currently, if the logged in user is already part of a group and the group has a supplier, when starting an order the user is still prompted to create a group / supplier as part of the pre-reqs. 

These prompts should be removed if the user is already part of at least 1 group, and / or if that group already has at least 1 supplier.

michael-smith-nz:
With this change, we'll need to think about how someone can create a group when one already exists. Where would that function exist? Related to #158

michael-smith-nz:
Hmmm, a few things unexpected happening here:

- the `hasNotCompletedGroupOrSupplierProfile` function on `orders.js` is using `sourceId` to find admin agents, but should be using `targetId`
- 2 new group agents are being created along with a bunch of relationships, etc. every time the 'Start Order' button is pressed

iainkirkpatrick:
the 2 group agents being created each time Start Order is clicked is to have a group and a supplier to populate as part of the pre-reqs. IMO this should only happen if the user isn't already part of a group, and if any of their groups don't have a supplier - i.e. only create these agents if the user needs to do that part of the pre-reqs

michael-smith-nz:
Is there any point having a 'Pre-Requisites' task? 
At the moment we have a complicated process of creating an empty agent and profile of a supplier and group and then updating them later down the pipeline. We've got a bunch of hooks in place to check if a group or supplier does not meet different criteria because of this. I suggest we move creation of groups and suppliers to the 'start order' task (mockup to come) and possibly somewhere else if users wanted to create groups and suppliers outside of starting an order.

michael-smith-nz:
Thanks to @iainkirkpatrick I've found Mikey working on #300. Looks like this will fix this issue as there will be no Pre-Requisites task.

michael-smith-nz:
Not relevant anymore

PR #300, opened 2017-12-05
wip: dash the boards

### LINKED ISSUE #298: intuit the ux
this ticket is to take a minimal viable set of changes from https://github.com/root-systems/cobuy/issues/157 that can make the ux more intuitive

ahdinosaur:
closes #298 

TODO

- [x] move tasks to separate navigable page
- [x] dumb components
  - [x] OrderList
  - [x] OrderListItem
  - [x] Order
  - [x] OrderCreator
- [x] add order creator page
- [x] fetch all related order data
- [x] and get orders as a higher-level data abstraction
- [x] add orders page (order list and button to open order creator)
- [x] add order page
- [x] replace dashboard page with orders page
- [ ] ???


michael-smith-nz:
Hey @ahdinosaur  This looks great! As mentioned on Slack and #290 it would be great if we could add creating a supplier and group to this list to remove the 'Pre-Requisites' task. 
Do you have any plans for this already? If not, I'll get onto it  😄

ahdinosaur:
>  it would be great if we could add creating a supplier and group to this list to remove the 'Pre-Requisites' task.
> Do you have any plans for this already? If not, I'll get onto it 😄

when creating an order, you can choose to create a new consumer or supplier with that order. if you do this, it should then create a pre-reqs task to fill in the details for that new consumer or supplier. this is my best take on a solution for #290, do you have a suggestion for something different than this?

danalexilewis:
@ahdinosaur what’s a consumer in the above? Is it a group?

In regards to the ux of this process I think a good analogy is toggl where you can quick create tags, clients, or projects from the filters. However if you want to set more refined details you need to go to the corresponding management page.

I think this is balancing act of **enabling** peeps to create new groups or suppliers (and products) without **derailing** the order creation process. 

So If a user wishes to add a new group or supplier to the order/system we could think of the tasks we give them in three categories:
 - those that have to be done as part of the order creation workflow, ie extending it. (Eg group name, supplier email)
 - those that have to be done before the order close (eg invite members)
 - those that are optional (eg add group photo)

Another angle we could take for groups is dividing between individual agent tasks and group tasks.

ahdinosaur:
the current progress is merged with #309, will deploy to production to start testing live.

PR #319, opened 2017-12-18
Dashboard only show current agents orders

### LINKED ISSUE #312: a hook on orders.find to limit orders by respective groups
currently everyone can see everything

michael-smith-nz:
Shows orders on the dashboard if the user is an admin or if the order is from a group they are a member of.

May still need to update query in container to fetch only related orders instead of all orders

fixes #312


gregorykan (approved):
👍

PR #320, opened 2017-12-19
bugfix(orders): query for supplierId if not done yet

gregorykan:


ordering/containers/OrderCreator.js
```diff
@@ -126,12 +126,23 @@ function OrderCreatorContainer (props) {
   return h(OrderCreator, nextProps)
 }
 
+const variadicEither = (head, ...tail) => reduce(either, head, tail)
+
+const hasQueriedForSupplierAgentId = (relationships) => {
```
ahdinosaur: has _not_ queried?

ahdinosaur: ramda party!

```js
const hasNotQueriedForSupplierAgentId = unless(isNil, pipe(
  map(path(['relationship', 'args', 'params', 'query', 'relationshipType', '$in'])),
  flatten,
  none(equals('supplier'))
))
```

ahdinosaur: less "pure" but still ramda:

```js
const hasNotQueriedForSupplierAgentId = (relationships) => {
  if (isNil(relationships)) return
  return pipe(
    map(relationship => relationship.args.params.query.relationshipType.$in),
    flatten,
    none(x => x === 'supplier')
  )(relationships) 
}
```

ahdinosaur: smells:

1. flatten(map(...)) - or any time you wrap the result of a function with another function. it's hard to tell what's happening in what order
2. map(fn, data) - or any time you call the function with data immediately, as opposed to creating a function on one line with mappedFn = map(fn) and calling that function with mappedFn(data)

ahdinosaur (commented):
🤶

PR #322, opened 2017-12-19
bugfixes: looping query on start order, tasks showing up as complete when they aren't

gregorykan:
- [x] fixed looping query on start order
- [x] fixed tasks showing up as complete when they aren't


ahdinosaur (approved):
:tophat:

PR #352, opened 2017-12-20
fix for app crash when collective intents quantity is less than all m…

### LINKED ISSUE #350: app crashing after trying to add intent
- sometimes price specs are not retrieved properly
- error being thrown from same getter, `ordering/getters/getCurrentOrderCollectiveQuantityByProduct.js`

iainkirkpatrick:
…inimums

closes #350


iainkirkpatrick:
we have user testing tomorrow, so gonna merge this 🏀

PR #383, opened 2018-01-08
feat(app): navbar logo links to root

### LINKED ISSUE #356: clicking tapin:buy should take you home

gregorykan:
closes #356


app/components/Navigation.js
```diff
@@ -59,7 +59,7 @@ function Navigation (props) {
       }
 
       return (
-        h(NavLink, {
+        h(Link, {
```
ahdinosaur: NavLink ensures the link can be highlighting, this is in the navigation column so should stay using NavLink

ahdinosaur (commented):
sweeeeeet! if we keep NavLink in the nav, the new Link is great!

PR #384, opened 2018-01-08
feat(orders): disable dropdown and use initial values for new consume…

### LINKED ISSUE #359: order creator without existing groups is confusing
- why have a dropdown when there's only one option?
- instead of 'NEW', say 'CREATE'
- perhaps include note to say that there are no groups yet
- MW: alternatively, auto-complete rather than dropdown

gregorykan:
…r and supplier if none exist yet

closes #359

### DIFF
--- a/ordering/containers/OrderCreator.js
+++ b/ordering/containers/OrderCreator.js
@@ -126,12 +126,13 @@ function OrderCreatorContainer (props) {
 
   const hasNoCurrentAgentGroupProfiles = isEmpty(props.currentAgentGroupProfiles)
 
+  // GK: for now we will use the 'new supplier/consumer' as default, will want to look later at making this more responsive
   const nextProps = merge(props, {
     onSubmit: handleSubmit,
     hasNoCurrentAgentGroupProfiles,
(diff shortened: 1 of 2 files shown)


ordering/containers/OrderCreator.js
```diff
@@ -124,8 +124,15 @@ function OrderCreatorContainer (props) {
...
   }
 
+  const hasNoCurrentAgentGroupProfiles = isEmpty(props.currentAgentGroupProfiles)
+
   const nextProps = merge(props, {
-    onSubmit: handleSubmit
+    onSubmit: handleSubmit,
+    hasNoCurrentAgentGroupProfiles,
+    initialValues: {
+      supplierAgentId: hasNoCurrentAgentGroupProfiles ? 'NEW_SUPPLIER' : '',
```
ahdinosaur: won't this always set the initial values to new supplier? since i think this code will always run first without having loaded the agents, which will permanently set the initial values.

it's probably fine to always default this to new supplier though, at least worth trying as a new test on users.

PR #393, opened 2018-01-09
bug(orderCreator): only close modal if valid

ahdinosaur:


ahdinosaur:
merging because done as a pair with @gregorykan

PR #394, opened 2018-01-09
hints everywhere!

### LINKED ISSUE #365: should have hints/guides everywhere on the app
- tooltips to start with
- this ticket is just to get us started

ahdinosaur:
partialfix for #365


ahdinosaur:
was thinking of using [font awesome icons](http://fontawesome.io/icons/) for "help":

screenshot_20180109_224642

this is what the proof-of-concept looks like, on hover over the icon:

screenshot_20180109_225238

/cc @agentlewis

danalexilewis:
Oh I see - yes I think this will work well. We can just put an info icon in the top.righr of every component - or something like that.

PR #402, opened 2018-01-11
deployment 110118 round 2

gregorykan:


ahdinosaur:
re-seeded the database with

```shell
heroku run npm run db seed:run --app cobuy-tapin
```

PR #403, opened 2018-01-11
WIP: Anywhere Order Summary

### LINKED ISSUE #389: should be able to see order summary at any time
especially admins are keen to see what the order summary would look like if sent to suppliers

michael-smith-nz:
Before this PR order summary uses `orderPlans` to show what users have ordered. `orderPlans` are only created when an order is closed. We use `orderIntents` until then.
`orderIntents` and `orderPlans` have the same data shape, so this PR should hopefully be as easy as switching `orderPlans with `orderIntents` when an order is still open.

Currently, this doesn't work for open orders. I suspect the problem to be in the `shouldQueryAgain` checks. There's also some renaming of functions that need to happen as most util functions now accept both `orderPlans` and `orderIntents`

fixes #389

### DIFF
--- a/ordering/components/OrderSummary.js
+++ b/ordering/components/OrderSummary.js
@@ -2,7 +2,7 @@ import { compose } from 'recompose'
 import { connect as connectFela } from 'react-fela'
 import h from 'react-hyperscript'
 import Paper from 'material-ui/Paper'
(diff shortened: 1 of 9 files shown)


michael-smith-nz:
Blocked on this. `status.isPending` is always returning true and not fetching priceSpecs, etc.. Might be bug with feathers-action-react?

PR #412, opened 2018-01-13
Feature/add name field to order

### LINKED ISSUE #390: add name to order
add name field to order creator

so we can display order name instead of order id

gregorykan:
closes #390


ahdinosaur (approved):
clean and clear, nice work @gregorykan! :cake:

PR #437, opened 2018-01-15
Feature/member invite emails sent on order start

### LINKED ISSUE #409: (Admin) Add the supplier and products before members
Currently when the members received their invites there was nothing in the order as the supplier and products had not been created yet. 

Rather we should add the supplier and products before this step so that everything is ready for when the members are invited.

### LINKED ISSUE #415: (email) invite call to action should take the user directly to the order they have been invited to
at the moment it takes the user to the landing page (currently the orders page)

### LINKED ISSUE #438: invite emails should only be sent after order has started
- i.e. after prereqs complete

duplicates #409

gregorykan:
closes #415 
closes #438 
closes #409


tasks/services/works.js
```diff
@@ -89,6 +94,112 @@ function createOrderPlans (hook) {
     .then(() => hook)
 }
 
+function sendOrderStartEmail (credential, appConfig, mailer, order) {
```
ahdinosaur: recommend using a single `options` argument rather than many positional arguments, so:

```js
function sendOrderStartEmail ({ credential, appConfig, mailer, order }) {
```

or

```js
function sendOrderStartEmail (options) {
  const {
    credential,
    appConfig,
    mailer,
    order
  } = options
}
```

ahdinosaur: recommend using a single `options` argument rather than many positional arguments, so:

```js
function sendOrderStartEmail ({ credential, appConfig, mailer, order }) {
```

or

```js
function sendOrderStartEmail (options) {
  const {
    credential,
    appConfig,
    mailer,
    order
  } = options
}
```

tasks/services/works.js
```diff
@@ -89,6 +94,112 @@ function createOrderPlans (hook) {
...
+
+      <br />
+      <br />
+
+      Click <a href=${appConfig.url}/o/${order.id}>here</a> to join the order!
+    `
+  })
+}
+
+function sendWelcomeEmail (credential, appConfig, mailer, order, token) {
```
ahdinosaur: options arg

tasks/services/works.js
```diff
@@ -89,6 +94,112 @@ function createOrderPlans (hook) {
...
+  const appConfig = hook.app.get('app')
+  const mailer = hook.app.service('mailer')
+
+  const hasPassword = (credential) => {
+    return not(isNil(credential.password))
+  }
+
+  return (credential) => {
+    if (hasPassword(credential)) {
+      return sendOrderStartEmail(credential, appConfig, mailer, order)
```
ahdinosaur: pass in options arg

tasks/services/works.js
```diff
@@ -89,6 +94,112 @@ function createOrderPlans (hook) {
...
+    } else {
+      return tokens.create({
+        agentId: credential.agentId,
+        service: 'credentials',
+        method: 'patch',
+        params: { serviceId: credential.id }
+      })
+      .then((token) => {
+        return sendWelcomeEmail(credential, appConfig, mailer, order, token)
```
ahdinosaur: pass in options arg

tasks/services/works.js
```diff
@@ -89,6 +94,112 @@ function createOrderPlans (hook) {
...
+    `
+  })
+}
+
+function sendEmailBasedOnPasswordStatus (hook, order) {
+  const tokens = hook.app.service('tokens')
+  const appConfig = hook.app.get('app')
+  const mailer = hook.app.service('mailer')
+
+  const hasPassword = (credential) => {
```
ahdinosaur: ramdy

```js
const hasPassword = pipe(prop('password'), isNil, not)
```

tasks/services/works.js
ahdinosaur: recommend using a single `options` argument rather than many positional arguments, so:

```js
function sendOrderStartEmail ({ credential, appConfig, mailer, order }) {
```

or

```js
function sendOrderStartEmail (options) {
  const {
    credential,
    appConfig,
    mailer,
    order
  } = options
}
```

tasks/services/works.js
```diff
@@ -89,6 +94,112 @@ function createOrderPlans (hook) {
...
+    return orders.get(taskPlanResult.params.orderId)
+    .then((orderResult) => {
+      return relationships.find({
+        query: {
+          sourceId: orderResult.consumerAgentId,
+          relationshipType: 'member'
+        }
+      })
+      .then((relationshipResults) => {
+        const agentIds = map((r) => r.targetId, relationshipResults)
```
ahdinosaur:
```js
const getTargetAgentIds = map(prop('targetId'))
const targetAgentIds = getTargetAgentIds(relationshipResults)
```

where the first line could be at the top scope

tasks/services/works.js
```diff
@@ -89,6 +94,112 @@ function createOrderPlans (hook) {
...
+        return credentials.find({
+          query: {
+            agentId: {
+              $in: agentIds
+            }
+          }
+        })
+        .then((credentialResults) => {
```
ahdinosaur:
```js
.then(pipe(
  map(sendEmail),
  Promise.all
))
```

where `sendEmail` can be defined in the scope above (after line 175)

tasks/services/works.js
```diff
@@ -89,6 +94,112 @@ function createOrderPlans (hook) {
...
+function sendStartOrderEmails (hook) {
+  const taskPlans = hook.app.service('taskPlans')
+  const orders = hook.app.service('orders')
+  const relationships = hook.app.service('relationships')
+  const credentials = hook.app.service('credentials')
+
+  return taskPlans.get(hook.data.taskPlanId)
+  .then((taskPlanResult) => {
+    return orders.get(taskPlanResult.params.orderId)
+    .then((orderResult) => {
```
ahdinosaur: can this be part of parent promise chain? doesn't seem like we need `taskPlanResult` again

tasks/services/works.js
```diff
@@ -24,6 +24,7 @@ const hooks = {
     create: [
       iff(taskRecipeIsCompleteOrderSetup, createCastOrderIntentTaskPlan),
       iff(taskRecipeIsCompleteOrderSetup, createCommitOrderTaskPlan),
+      iff(taskRecipeIsStartOrder, sendStartOrderEmails),
```
ahdinosaur, line 27: name of `sendStartOrderEmails` is confusing, you also have `sendOrderStartEmail`

ahdinosaur (commented):
here's a spray of comments :fountain: :love_letter:

tasks/services/works.js
gregorykan: refactor to `createStartOrderEmail`

tasks/services/works.js
gregorykan: dealt with this by changing names of the lower order functions~

ahdinosaur (approved):

PR #454, opened 2018-01-18
Tests/cobuying algorithms

### LINKED ISSUE #452: write tests for cobuying algorithms

iainkirkpatrick:
coverage of the getters used in the two applications of the buying algorithm (intents -> plans, and displaying current state of group intents on the client)

interested in thoughts on:
- are equality tests the ideal way to go about this? could try and write tests that aren't specific to the mock data used, seems tricky / fraught though...
- should the mock data be more complex for more thorough testing?

this is s start for now at least, makes it easier to protect / refactor these getters if need be

closes #452

### DIFF
--- a/ordering/data/mock.js
+++ b/ordering/data/mock.js
@@ -1,0 +1,43 @@
+export const mockOrderIntents = {
+  "1": {
+    "id": 1,
+    "agentId": 4,
+    "desiredQuantity": 5,
+    "productId": 1,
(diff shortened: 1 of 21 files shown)


ahdinosaur:
does this test all the "unhappy paths", all the possible ways to break the algorithm?

gregorykan (approved):
:+1: will create a further ticket to address @ahdinosaur 's concerns

PR #485, opened 2018-01-24
Bugfix/inviting existing user breaks everything

### LINKED ISSUE #347: inviting existing users will result in infinite loop
consider alongside #472

gregorykan:
closes #347 
relates to https://github.com/root-systems/dogstack-agents/pull/24

### DIFF
--- a/package.json
+++ b/package.json
@@ -72,7 +72,7 @@
     "deep-extend": "^0.5.0",
     "dog-names": "^1.0.2",
     "dogstack": "^0.7.0",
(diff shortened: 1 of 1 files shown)


iainkirkpatrick:
@gregorykan this doesn't need to happen i don't think now that we have merged the fix in `dogstack-agents` in ? i.e. once `dogstack-agents` is re-published, we can just reinstall that module locally / on prod and it will update

gregorykan:
should not need these changes once new `agents` has been published

PR #486, opened 2018-01-24
feat(profile): added route and container for profile WIP

### LINKED ISSUE #456: my profile route/page
- should show buying groups
- top-right in navbar
- in navbar also

gregorykan:
closes #456 

- [x] new route
  - [x] add link to navbar
- [x] new container for my-profile
- [x] show groups i belong to

N.B. the save/edit functionality/flow will need to be clarified at some point. not a block right now as moving on to work on the other profile types

### DIFF
--- a/agents/components/Profile.js
+++ b/agents/components/Profile.js
@@ -1,6 +1,6 @@
 import { connect as connectFela } from 'react-fela'
 import { Field, reduxForm as connectForm } from 'redux-form'
-import { pipe, isNil, not } from 'ramda'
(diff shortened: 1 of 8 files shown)


iainkirkpatrick (approved):
looks good! only thing i can see is pre-populating the profile email with the email entered when signing up... but this is an issue with dogstack-agents, i'll open a separate ticket

also in future i guess we'll want to do something with the groups rather than just a list? group profile pages? can fix then

PR #492, opened 2018-01-31
Chore/harden services

### LINKED ISSUE #460: feathers service permissions/authentication

iainkirkpatrick:
Adds auth hook to all relevant services, plus more thought-out hooks + tests to services:
- credentials (cobuy additional hooks)
- notifications
- orderIntents
- orderPlans
- orders
- resourceTypes
- priceSpecs

None of the tests yet simulate an authorised client connecting to the server and calling services - i haven't yet got a test client + server combo communicating properly. so at the moment any tests which need params which include `{ provider: 'rest' }` won't run unless the `all: authenticate('jwt')` is turned off first.

closes #460


lib/hooks/restrictToCurrentUserGroups.js
```diff
@@ -0,0 +1,56 @@
...
+    if (!hook.params.query.consumerAgentId) {
+      hook.params.query = merge(hook.params.query, {
+        consumerAgentId: {
+          $in: groupIds
+        }
+      })
+      return hook
+    }
+    // either a number, or an object with querying syntax
+    if (typeof hook.params.query.consumerAgentId === 'object') {
```
gregorykan, line 38: would you consider using `path` here as this seems fragile?

gregorykan: either path or destructuring earlier?

lib/hooks/restrictToCurrentUserGroups.js
```diff
@@ -0,0 +1,56 @@
...
+      if (hook.params.query.consumerAgentId.$in) {
+        const disallowedGroups = difference(hook.params.query.consumerAgentId.$in, groupIds)
+        hook.params.query.consumerAgentId.$in = without(disallowedGroups, hook.params.query.consumerAgentId.$in)
+        return hook
+      }
+      hook.params.query.consumerAgentId = merge(hook.params.query.consumerAgentId, { $in: groupIds })
+      return hook
+    }
+    if (!contains(hook.params.query.consumerAgentId, groupIds)) {
```
gregorykan, line 47: yeah you use this path so much, may be worth declaring/destructuring earlier

iainkirkpatrick: had a wee look into it now, think it's not worth doing for now... given the spots where we need to set a property on the query, which we can't do with a declared `path` - it would be a mix of `path` and the full dot syntax path anyway. gonna merge this and we can come back to it later in some other refactor 🙏

